PR #5, opened 2014-01-29
Gaelmuller syscheck improvemnts

jrossi:
Several improvements / corrections in syscheck:
- Report file deletion, even without realtime enabled
- Report modifications made on directories
- Corrects bug so that files created between the first and second scan are reported as new files
- Corrects bug that made changes reverting a file to the state it was in when ossec started unreported
- Avoids computing hashes multiple times to improve performance
- Make the time between two syscheck wakeups configurable in internal_options

Orginal Pull Request: https://bitbucket.org/jbcheng/ossec-hids/pull-request/19/several-improvements-corrections-in/diff

@gaelmuller

### DIFF
--- a/src/syscheckd/run_check.c
+++ b/src/syscheckd/run_check.c
@@ -392,17 +392,16 @@ void start_daemon()
 
                 /* Checking for changes */
                 run_dbcheck();
-            }
 
+                /* Sending scan ending message */
+                sleep(syscheck.tsleep + 20);
+                if(syscheck.dir[0])
+                {
+                    merror("%s: INFO: Ending syscheck scan.", ARGV0);
+                    send_rootcheck_msg("Ending syscheck scan.");
 
-            /* Sending scan ending message */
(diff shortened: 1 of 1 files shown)


chrystalis:
At least one item in here is still an issue for me when running syscheck with realtime enabled: "Corrects bug that made changes reverting a file to the state it was in when ossec started unreported"

I'm not a C coder, so I can't judge whether this patch was good to be included or not.  If not, I'll submit a new issue for it, as it does seem to me to be a bug.

apobereyko:
While testing ossec-2.9.1 with an agent on Windows 10 platform I still experience problem with removed files not detected on Windows machine. Adding/modifying files does work just fine.

Debug log shows only 1 event for a file, so not sure how to actually see all events related to a file and track modification history.

Any help/hint would be valuable.

Thanks in advance.

billy-shears:
I am having the same problem as in previous comment. Deletions on monitored file systems are not detected.

PR #438, opened 2014-11-15
LIBMAGIC Has not been added into the new Makefile system. 

jrossi:
Bring back the libmagic option for some sytems.

```
make TARGET=server USE_MAGIC=auto 
```


awiddersheim:
Looks like this needs another merge from master but seems good to me.

jrossi:
It's hidden but this is f-ed up.  I have been working on it in my repo and this change has becomes massive and I don't think libmagic has worked in a long time with ossec :)

Issue #224, opened 2014-06-08
Windows Events Not Being Decoded With Eventchanel


mstarks01:
When using eventchannel over eventlog, no Windows alerts will be logged in alerts.log and alerts sent. So, basically, it's not usable.

The issue seems to be that the log is sent in a multi-line format. From archives.log:

```
2014 Jun 08 10:20:16 (promenade) 192.168.1.12->WinEvtLog 2014 Jun 08 10:14:47 WinEvtLog: Security: Information(4732): no source: no user: no domain: promenade: A member
 was added to a security-enabled local group.

Subject:
        Security ID:            S-1-5-21-4082012827-2132021701-1268594856-1002
        Account Name:           me
        Account Domain:         promenade
        Logon ID:               0x2926e5b

Member:
        Security ID:            S-1-5-21-4082012827-2132021701-1268594856-501
        Account Name:           -

Group:
        Security ID:            S-1-5-32-544
        Group Name:             Administrators
        Group Domain:           Builtin

Additional Information:
        Privileges:             -
```

When using eventlog as the format, the line looks like this:

```
2014 Jun 08 11:10:57 (promenade) 192.168.1.12->WinEvtLog 2014 Jun 08 11:06:11 WinEvtLog: Security: AUDIT_SUCCESS(4732): Microsoft-Windows-Security-Auditing: (no user): no domain: promenade: A member was added to a security-enabled local group. Subject:  Security ID:  S-1-5-21-4082012827-2132021701-1268594856-1002  Account Name:  me  Account Domain:  promenade  Logon ID:  0x2926e5b  Member:  Security ID:  S-1-5-21-4082012827-2132021701-1268594856-501  Account Name:  -  Group:  Security ID:  S-1-5-32-544  Group Name:  Administrators  Group Domain:  Builtin  Additional Information:  Privileges:  -
```

And everything works normally.

PR #457, opened 2014-12-01
Fix windows event channel

awiddersheim:
Quite a few changes/fixes/updates to the Windows event channel code. Many changes were done to help address issues in https://github.com/ossec/ossec-hids/issues/224. Some testing was done by myself and @defensivedepth but more testing would certainly be appreciated. Need people who use Windows and event channel.

Tried to make the output from the `eventlog` format and the `eventchannel` format as similar as possible. Added a lot of error checking. Fixed issues where not all log messages were able to be displayed. Moved to using temporary files and atomic moves to update bookmarks.

### DIFF
--- a/src/logcollector/logcollector.c
+++ b/src/logcollector/logcollector.c
@@ -812,21 +812,11 @@ void win_format_event_string(char *string)
 
     while(*string != '\0')
     {
-        if (*string == '\n' || *string == '\r')
+        if (*string == '\n' || *string == '\r' || *string == ':')
         {
-            *string = ' ';
-            string++;
-
-            while(*string == '\t')
-            {
+            if (*string == '\n' || *string == '\r')
                 *string = ' ';
-                string++;
-            }
 
-            continue;
-        }
-        else if (*string == ':')
-        {
             string++;
 
             while(*string == '\t')


defensivedepth:
Just an update on this... I am noticing a much higher than average ossec client memory usage on the 2008 R2 servers using this build... A couple of them are 500mb up to 800mb, with only an up-time of 2 days...

awiddersheim:
Hrm, is the memory rising or does it stay at that level? What are "normal" levels when it comes to OSSEC on Windows? 

Sounds like a memory leak. Question is, is the leak in the eventchannel code or somewhere else. The build you are using has not only my changes but quite a few others that MAY be causing issues. Any errors in the OSSEC log? If you disable eventchannel and restart the agent does the memory issue persist?

defensivedepth:
After looking into this further, It appears to gradually increase after a reboot until all memory is consumed on the box...  

Normal levels that I have typically observed is under 10mb...  

No errors in the OSSEC log.

I will disable event channel, restart the agent and get back to you on if the issue persists...

awiddersheim:
Definitely seems like a memory leak. I see a few issues with the event channel code that might fix the issue but might take me a bit before I get enough free time to work on it. Let me know how your testing goes with disabling event channel.

defensivedepth:
After disabling the eventchannel logs and restarting the agent, memory usage sat at under 10mb for the next 5 hours. 

Once I enabled the eventchannel logs and restarted the agent, I saw an instant CPU & memory spike (which is to be somewhat expected, as it shipped off the 50k events that were generated and not sent to the OSSEC server during those 5 hours).. During this time when the new events were being processed, the memory usage of the agent went from <10mb to 127mb, and then the memory increase became much slower, correlating to when the agent stopped processing the backlog.  

During the next hour, memory usage continued to increase, but more slowly... Currently at 155mb and increasing.

Based on this testing, I would agree that the leak appears to be related with the eventchannel code. 

No worries on the timing... If I have to have this functionality right now, I can just set the agent to restart hourly...

awiddersheim:
Great info and testing notes. Thanks for testing and helping to narrow this down.

defensivedepth:
Let me know if I can do any more testing, etc

awiddersheim:
Found that this buffer never has `free()` called.

https://github.com/ossec/ossec-hids/blob/master/src/logcollector/read_win_event_channel.c#L572

This is probably the biggest source of memory leakage that I created.

@jrossi also pointed out several spots where I should be doing an `EvtClose()` on event handles. Should have something shortly that will hopefully fix things.

awiddersheim:
Most of the memory leaks were fixed in https://github.com/ossec/ossec-hids/pull/496. Turns out I didn't introduce the memory leaks after all. Most were already there I just didn't see them.

PR #471, opened 2014-12-23
Add bool to rootcheck and syscheck to skip NFS mounted partitions

reyjrar:
This is only implemented for Linux and FreeBSD as they are the only OS's I can track down decent documentation on how to accomplish this.  If this variable is set to "yes" and the feature isn't availble, a verbose error message is output to the OSSEC log file and the check returns false so the calling function will scan the directory as normal.

Need to document the change in ossec-docs, but basically adds a

    <skip_nfs>no</skip_nfs>

To the rootcheck and syscheck configurations.  To disable scanning of NFS shares:

    <skip_nfs>yes</skip_nfs>

Comments?


jrossi:
@reyjrar can you take a look into the comments above?

reyjrar:
It's a good idea, I like it!  I just haven't had the chance to add it yet.  I'll look at that tomorrow.

reyjrar:
Bailing on this, attempting to create another pull request.

PR #476, opened 2015-01-05
Add new rule to proftpd ruleset

ChristianBeer:
Add rule tests for proftpd log samples (added date/timestamp to samples from decoder.xml)


ddpbsd:
Thanks!

PR #478, opened 2015-01-05
Add a -F to read log messages from a file.

ddpbsd:
Might be useful for testing.


jrossi:
Clang does not like this 

```
analysisd/testrule.c:126:19: error: expression which evaluates to zero treated
      as a null pointer constant of type 'char *'
      [-Werror,-Wnon-literal-null-conversion]
    char *lfile = '\0';
                  ^~~~
1 error generated.
```

ddpbsd:
Can't figure out clang.

PR #479, opened 2015-01-05
Minor: When testing php files use \S+ not \w+ for filename pattern

dangarthwaite:


dangarthwaite:
Deleting - This PR pulled in gentests.py code and it is inappropriate.

PR #486, opened 2015-01-07
Test web appsec rules

dangarthwaite:
An initial, passing set of rules for web_appsec_rules.xml.

There are a  few rules that are commented out because I know of now way of testing multiple similar log entries without rewriting runtests.py.

Some caveats and observations:
As tests these are pretty useless without corresponding False Positive testing log entries.  

Many of the requests were written as POSTs because of a precompiled URL check for 'basic requests' that prevent things like User Agent string matching later.

I also reduced the Alert level on both the test and the rule for a Wget user-agent from 6 to 0.


ddpbsd:
Thanks!

Issue #463, opened 2014-12-16
segfault in analysisd caused by rule level change


mjeffe:
I recently upgraded OSSEC from version 2.7.1 to 2.8.1. The new version of analysisd kept segfaulting.  After a couple of weeks of working with it, I've narrowed it down to a simple rule override in `local_rules.xml` where I downgrade the level. Below I describe my test scenario where I can consistently reproduce the segfault with a bare minimum.

Current environment is Amazon AWS VPC. All servers are running Amazon Linux 64-bit AMI's (based on RHEL/CentOS).
```
[root@ossectst ossec]# uname -a
Linux ossectst 3.14.23-22.44.amzn1.x86_64 #1 SMP Tue Nov 11 23:07:48 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux
```

Here is what I did:

1) I spun up a brand new server (clone of my current ossec server), installed a fresh 2.8.1 ossec server. Modified local_rules.xml to look like this:

```
<!-- @(#) $Id: ./etc/rules/local_rules.xml, 2011/09/08 dcid Exp $
<group name="local,syslog,">
  <!-- downgrade from level 7 to level 5 -->
  <rule id="533" level="5" overwrite="yes">
    <if_sid>530</if_sid>
    <match>ossec: output: 'netstat -tan</match>
    <check_diff />
    <description>Listened ports status (netstat) changed (new port opened or closed).</description>
  </rule>
</group> <!-- SYSLOG,LOCAL -->
```
I did NOT add any agents, but let that run for about an hour (may not be necessary).

2) I then stopped the ossec server, ran manage_agents, added one agent and restarted the ossec server. NOTE, I did not transfer the agent key, so the agent was not trying to communicate yet. I let that run for about an hour (again, may not be necessary).

3) I transferred the key to the agent - this is an existing agent, I just added the new key and change the `<server-ip>` to point to my new test server and restarted the agent and server. I begin to see communication between server and agent. I let it run it's initial scans - about 10 minutes.

4) Then I opened a new listen port to try and trigger the `netstat -tan` rule using `nc -l 7777` on the agent.

5) ossec-analysisd segfaulted after about 7 minues.

Let me know if you want any of the output or if you want me to run any other tests.

[...]

PR #494, opened 2015-01-12
Fix for issue #463 (rule overwrites causing a segfault)

ddpbsd:
This causes the <check_diff /> option to not crash overwrites of sid 533 (issue #463). It would require EXTENSIVE testing, or someone who knows
what they are doing to look it over.

### DIFF
--- a/src/analysisd/analysisd.c
+++ b/src/analysisd/analysisd.c
@@ -1662,8 +1662,7 @@ RuleInfo *OS_CheckIfRuleMatch(Eventinfo *lf, RuleNode *curr_node)
     if(currently_rule->context == 1)
     {
 
-	if(currently_rule->context_opts & SAME_DODIFF) {
-	} else {
+	if(!(currently_rule->context_opts & SAME_DODIFF)) {
         	if(!currently_rule->event_search(lf, currently_rule))
             	return(NULL);
 	}
--- a/src/analysisd/analysisd.c
+++ b/src/analysisd/analysisd.c
@@ -1661,8 +1661,12 @@ RuleInfo *OS_CheckIfRuleMatch(Eventinfo *lf, RuleNode *curr_node)
     /* If it is a context rule, search for it */
     if(currently_rule->context == 1)
     {
-        if(!currently_rule->event_search(lf, currently_rule))
-            return(NULL);
+
+	if(currently_rule->context_opts & SAME_DODIFF) {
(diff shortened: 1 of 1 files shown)


src/analysisd/analysisd.c
```diff
@@ -1661,8 +1661,12 @@ RuleInfo *OS_CheckIfRuleMatch(Eventinfo *lf, RuleNode *curr_node)
...
     if(currently_rule->context == 1)
     {
-        if(!currently_rule->event_search(lf, currently_rule))
-            return(NULL);
+
+	if(currently_rule->context_opts & SAME_DODIFF) {
+	} else {
+        	if(!currently_rule->event_search(lf, currently_rule))
+            	return(NULL);
+	}
```
jrossi: Would this not amke more sense writen like this: 

```C
if !(currently_rule->context_opts & SAME_DODIFF) {
     if(!currently_rule->event_search(lf, currently_rule))
     return(NULL);
}
```
 Or am i missing something?

ddpbsd: @jrossi That definitely seems cleaner than what I have. Updating now.

src/analysisd/analysisd.c
```diff
@@ -1661,8 +1661,11 @@ RuleInfo *OS_CheckIfRuleMatch(Eventinfo *lf, RuleNode *curr_node)
     /* If it is a context rule, search for it */
     if(currently_rule->context == 1)
     {
-        if(!currently_rule->event_search(lf, currently_rule))
-            return(NULL);
+
+	if(!(currently_rule->context_opts & SAME_DODIFF)) {
+        	if(!currently_rule->event_search(lf, currently_rule))
+            	return(NULL);
```
jrossi: i am being to pedantic in that this indention is very unclear.   Going to pull this in, as we have not formalized the coding sytles stuff yet.

ddpbsd: It does look off. I must have fudged something earlier.

jrossi: This is personal choice I hate bare if blocks as they get hard to read.  I know it more typing but what i did here is always more clear to me when reading: 

```
if(!(currently_rule->context_opts & SAME_DODIFF)) {
    if(!currently_rule->event_search(lf, currently_rule)){
        return(NULL)
    }           
}
```

PR #497, opened 2015-01-14
Fix the wrong ARGV0 defined for authd causing the init script to fail

nixfloyd:
Fix the wrong ARGV0 defined for authd (which caused, among other things, the pidfile to be written to '.../auth-${pid}.pid' instead of '.../ossec-authd-${pid}.pid'


jrossi:
Looks great will merge once Travis confirms it all builds and tests.

nixfloyd:
Thanks @jrossi

PR #501, opened 2015-01-21
Add support for FQDN in csyslogd

ccooke:
Hi,

The existing csyslogd always truncates hostnames in the syslog output at the first '.', which creates a lot of ambiguity if you differentiate hosts by a hierarchical DNS scheme. These patches add a "use_fqdn" option into the syslog config that toggles between using the short or full hostname. 

Some notes:
* The default behaviour is to use the short hostname, so current behaviour is preserved
* I would have preferred to make this a global option (since hostnames are truncated in a few places) but it looks like csyslogd only has infrastructure to read its own configuration at the moment. Adding the global option would be a much more invasive change
* The code works on a per syslog config block basis, which looks to be the expected method.

Thanks,

### DIFF
--- a/src/config/csyslogd-config.c
+++ b/src/config/csyslogd-config.c
@@ -30,6 +30,7 @@ int Read_CSyslog(XML_NODE node, void *config, __attribute__((unused)) void *conf
     const char *xml_syslog_id = "rule_id";
     const char *xml_syslog_group = "group";
     const char *xml_syslog_location = "location";
+    const char *xml_syslog_use_fqdn = "use_fqdn";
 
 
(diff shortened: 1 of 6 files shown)


ddpbsd:
I'd love to see a pull request for ossec-docs as well.

ccooke:
No problem. I have that written here - just checking, should I add to docs/syntax/head_ossec_config.syslog_output.rst or docs/syntax/ossec_config.syslog_output.trst 

Presuming the head version?

ddpbsd:
docs/syntax/ossec_config.syslog_output.trst please

PR #504, opened 2015-01-26
Ossec/master

hyn172:
This change is [Review on Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/504)


etc/rules/msauth_rules.xml
```diff
@@ -876,7 +876,13 @@
     <description>User Logoff Exchange.</description>
   </rule>
   
-  
+    <rule id="18262" level="3">
+    <if_sid>18107</if_sid>
+    <id>^4624$</id>
+    <match>Logon Type:   2</match>
```
jrossi: I would think this would be better if you matched with something like:

```
<match>Logon Type:\W+2 </match>
                  ^   ^-notice the space
                   \--- Anything other then A-Z, a-z, 0-9, '-', '@'
```

The reason is that this will match 2, 211, 21 just about anything starting with a 2

jrossi:
Could you also include an example of the log line this matches so that we know all the details of what is going on.  

If you look at https://github.com/ossec/ossec-hids/tree/master/contrib/ossec-testing we also have an testing framework for logs.  Might want to add this into that so regressions do not happen if other things change around this log line.

etc/rules/msauth_rules.xml
hyn172: yes, it's 3 spaces, default by microsoft win event. And i see windows logon event only maxium logon type =11

hyn172:
@jrossi : sorry, but can't guide me use ossec-testing? I offer only use ossec-logtest on server

jrossi:
Sorry busy week.   Ossec testing is basiclly just an INI file is contrib/ossec-testing/tests/ that has a log line and the expected output.  Should be simple to do, but to run them you must have python on the system as it calls into ossec-logtest to do it's work. 

You can call it form the Makefile `cd src &&  make test-rules` or directly with: 

```
 cd contrib/ossec-testing && sudo python runtests.py
```

jrossi:
@hyn172 does the ossec-testing stuff make sense?  Anything else you were wondering about after looking at the files (if you had time, which is hard to come by i know).

ChristianBeer:
@hyn172 can you provide some log samples that would trigger your new rule? I could write the ossec-testing stuff so this can be included to 2.9.

hyn172:
Sorry, my country, we have spent tet holiday in two weeks. So, next week,
i'll provide to you some log samples (my test system at office)

On Sat, Feb 14, 2015 at 2:26 AM, Christian Beer <notifications@github.com>
wrote:

> @hyn172 <https://github.com/hyn172> can you provide some log samples that
> would trigger your new rule? I could write the ossec-testing stuff so this
> can be included to 2.9.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ossec/ossec-hids/pull/504#issuecomment-74311237>.
>



-- 
--------------------------------------------------------
 Thanks & Best Regards.

awiddersheim:
@hyn172 it's been a while on this. We are working on getting 2.9 ready. Are you able to provide some log samples?

jrossi:
@hyn172 as we don't have anything to test this with and it's effect on other rules could be an issue we are going need some sample before we can merge.  

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ossec/ossec-hids/504#-:-KCg-2AU7rxzV0W-uaB7:1176294253)*

Issue #472, opened 2014-12-24
IIS 7/8 logs not triggering on Injection Attack Web Rule if response code matches ^2 or ^3


jameswhittington3:
> **A pattern of a SQL injection attack contained in IIS W3C Extended Logs with status code of 200 get caught by web rule  31108 (Ignored URLs (simple queries)).**

```
<rule id="31108" level="0">
      <if_sid>31100</if_sid>
       <id>^2|^3</id>
      <compiled_rule>is_simple_http_request</compiled_rule>
      <description>Ignored URLs (simple queries).</description>
   </rule>
```

> **I would expect it to be caught by rule 31103 (SQL injection attempt.) but it doesn't**

```
`<rule id="31103" level="6">
    <if_sid>31100</if_sid>
    <url>=select%20|select+|insert%20|%20from%20|%20where%20|union%20|</url>
    <url>union+|where+|null,null|xp_cmdshell</url>
    <description>SQL injection attempt.</description>
    <group>attack,sql_injection,</group>
  </rule>`
```

> **Below is a injection sample from a IIS 7 website:**

```
[...]

PR #508, opened 2015-01-27
enable web attack detection for IIS with status code 200

ChristianBeer:
This PR is related to #472 it needs to be tested some more with other log samples that trigger web_rules.xml so please don't merge this now.

If someone can supply me with some more samples (especially IIS ones) I willput them in this PR.


ChristianBeer:
@jrossi Can you take a look at the travis log? Why is the errorcode 3 if the rule, level and decoder are correct?

ChristianBeer:
@jrossi Found the flaw in the test. This is ready to be merged now.

Issue #327, opened 2014-09-30
Commit 71519bd7da1e3da757122ef47bd933e8fd133d62


cgzones:
Can someone explain to me, why these changes 71519bd7da1e3da757122ef47bd933e8fd133d62 (the os_regex related ones) are necessary, as the configuration language for ossec is English?
What is the purpose of the character -27 = 229 (http://www.theasciicode.com.ar/extended-ascii-code/capital-letter-o-tilde-ascii-code-229.html)

PR #511, opened 2015-01-28
remove unknown code

cgzones:


jrossi:
wow - @dcid normally did not add weird things like this into code.  I wonder what this was for.......

cgzones:
i already asked for the purpose of this code in #327, but no answer was given, so ...

ddpbsd:
@cgzones Did you ask dcid?

jrossi:
@ddpbsd i think i did as one point - will need to check into email :) don't think he remembered.

PR #512, opened 2015-01-29
exit on single unit test failure

cgzones:


wclarie:
Does that mean that if there are multiple checks, this will fail on the first one? Isn't is useful to find out all the failing tests instead of just the first one?

cgzones:
Good point. There are several binaries with several checks. Now all binaries are executed but if any fails the failure is not propagated to the makefile (so travis will succeed).
With this change make will abort after the first failed binary, but in this binary all checks are executed.

jrossi:
Seeing all the failures is useful, but passing tests becuase it was not the last one is a bug.  I am going to accept this and make sure things don't slip by we can make it better anytime.  

Plus this is how most of makefile currently works.  If it exits with anything other then 0 everything stops ;)

PR #514, opened 2015-01-29
fix warnings

cgzones:
fix several gcc warnings:
-Wabi -Wdouble-promotion -Winit-self -Wmissing-include-dirs -Wunused -Wsuggest-attribute=noreturn -Wsuggest-attribute=format -Wfloat-equal -Wundef -Wshadow -Wbad-function-cast -Wcast-qual -Wcast-align -Wlogical-op -Waggregate-return -Wstrict-prototypes -Wold-style-declaration -Wredundant-decls -Wnested-externs -Winline -Wc++-compat -Winvalid-pch -Wswitch-enum

### DIFF
--- a/src/analysisd/analysisd.c
+++ b/src/analysisd/analysisd.c
@@ -28,6 +28,7 @@
 #include "eventinfo.h"
 #include "accumulator.h"
 #include "analysisd.h"
+#include "fts.h"
 
 #ifdef PICVIZ_OUTPUT_ENABLED
 #include "output/picviz.h"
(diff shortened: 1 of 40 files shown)


src/analysisd/analysisd.c
```diff
@@ -1061,157 +1053,157 @@ RuleInfo *OS_CheckIfRuleMatch(Eventinfo *lf, RuleNode *curr_node)
      * weekday,
      * status,
      */
-    RuleInfo *currently_rule = curr_node->ruleinfo;
+    RuleInfo *rule = curr_node->ruleinfo;
```
jrossi: Does make this much easier to read :)

cgzones: this change was caused by -Wshadow, because there is a global variable named like that in https://github.com/ossec/ossec-hids/blob/master/src/analysisd/rules.h#L169

PR #515, opened 2015-01-29
fix some Coverity issues

cgzones:


awiddersheim:
Nice to have @cgzones back!

PR #526, opened 2015-02-02
possible fix to vista_sec.csv

### LINKED ISSUE #525: Fix vista_sec.csv
Fix formatting issue with https://github.com/ossec/ossec-hids/blob/master/src/win32/vista_sec.csv

image

ChristianBeer:
try to fix #525 I couldn't find the correct descriptions of those event ids. Maybe it's better to just delete them? I don't know how this is used inside OSSEC.


awiddersheim:
Delete the ones you put on their own lines?

awiddersheim:
This reads in the file creating a hash:

https://github.com/ossec/ossec-hids/blob/master/src/logcollector/read_win_el.c#L520

Then this searches the hash using the event ID and formats the message:

https://github.com/ossec/ossec-hids/blob/master/src/logcollector/read_win_el.c#L520

awiddersheim:
Anyway... this seems reasonable to me. I guess I'm still confused on your doubts about it.

ChristianBeer:
This will work for OSSEC but is still not real CSV and github will still complain about it.

awiddersheim:
Ah, because of the extra comma separators? Maybe we should rename the file so github won't think it's CSV anymore and properly handle indexing it and what not.

awiddersheim:
Or maybe use a different separator?

ChristianBeer:
I don't think it's useful to change the separator or the logic so it could read proper CSV so one file can be indexed by github.

As it stands now there was a bug in this file (improper description lines) that is now gone.

awiddersheim:
Yeah, but the work to change isn't outrageous which is why I proposed it. Also, I use git's search a lot and ran up against and issue where I was trying for stuff in this file and wasn't able to because git wasn't indexing I guess.

PR #530, opened 2015-02-03
Remove old build files that are no longer used

awiddersheim:
Now that the Windows builds have moved over to make we can get rid of a
lot of the files used for the old system. Also, remove files used for
building on Windows because that is no longer a supported/maintained
option at this point in time. That is not to say we won't accept any
patches from anyone willing to work on making it an option.

To quote jrossi... "Code wins"

Finally, remove the old favicon that doesn't seem to be used anymore.


cgzones:
i think the file `iis-logs.bat` and the build scripts in `win32/ui/` can also be deleted

awiddersheim:
@cgzones thanks. Let me take a look at those as well and I'll update this pull request.

PR #537, opened 2015-02-05
fix most gcc -Wall -Wextra warnings in windows build

cgzones:
There are 2 issues left:
- the mingw-gcc version used at travis does not like the initialization at https://github.com/ossec/ossec-hids/blob/master/src/logcollector/read_win_event_channel.c#L667, but newer versions do not complain
- on windows sockets are unsigned, so calls like https://github.com/ossec/ossec-hids/blob/master/src/client-agent/receiver-win.c#L53 causes a warning about sign/unsigned comparison. The hole os_net code does not take this into account, so that might be fixed later

### DIFF
--- a/src/rootcheck/rootcheck-config.c
+++ b/src/rootcheck/rootcheck-config.c
@@ -58,8 +58,9 @@ int Read_Rootcheck_Config(const char *cfgfile)
     const char *(xml_rootkit_winmalware[]) = {xml_rootcheck, "windows_malware", NULL};
     const char *(xml_scanall[]) = {xml_rootcheck, "scanall", NULL};
     const char *(xml_readall[]) = {xml_rootcheck, "readall", NULL};
+#ifdef OSSECHIDS
     const char *(xml_time[]) = {xml_rootcheck, "frequency", NULL};
-
+#endif
     const char *(xml_check_dev[]) = {xml_rootcheck, "check_dev", NULL};
     const char *(xml_check_files[]) = {xml_rootcheck, "check_files", NULL};
     const char *(xml_check_if[]) = {xml_rootcheck, "check_if", NULL};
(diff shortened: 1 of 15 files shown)


src/rootcheck/rootcheck-config.c
```diff
@@ -58,8 +58,9 @@ int Read_Rootcheck_Config(const char *cfgfile)
     const char *(xml_rootkit_winmalware[]) = {xml_rootcheck, "windows_malware", NULL};
     const char *(xml_scanall[]) = {xml_rootcheck, "scanall", NULL};
     const char *(xml_readall[]) = {xml_rootcheck, "readall", NULL};
+#ifdef OSSECHIDS
```
jrossi: Was not expecting to see OSSECHIDS come back

jrossi: Is define still needed in windows ? Thought we killed it in Unix

cgzones: i don't know, there is still the binary ossec-rootcheck.exe, which is bundled by the nsi script, but i do not know if somebody uses it on windows systems as standalone app.
Although @wclarie suggested to remove the standalone rootcheck source.

jrossi: I am all for killing it going forward :) we killed it on Unix.

awiddersheim: I'm not sure if anyone uses it either. Maybe a quick post to the non dev mailing list asking if anyone uses it? Depending on the response we go from there.

Seems like a pretty safe route to go. Gives people the opportunity to save it or gives reasons as to what they use it for, etc.

jrossi:
We can kill ossec-rootcheck.exe in another pull request.

Issue #442, opened 2014-11-17
ossec-logcollector stops reading alerts.log


ddpbsd:
It can stop after a minute or an hour.

More info can be found here: http://marc.info/?t=141503755300002&r=1&w=2

ddpbsd:
I think I've found the root of the problem. I have committed a potential fix here: https://github.com/ddpbsd/ossec-hids/commit/c6e63dabd2a2b11b5896bebbb620065dae299605 Please test it!

PR #540, opened 2015-02-05
Possible fix for Issue #442: logcollector stops reading alerts.log

ddpbsd:
I think this needs a decent amount of testing before it's included (unless someone works out the logic and I happen to be correct), but this seems to have fixed the issue of logcollector ignoring alerts.log

r usually does not equal 0, and this does not appear to be a parsingerror. logff[i].ign was being ++'d at that point every loop through, and is later compared to open_file_attempts. If you raise ign every time it's going to quickly be higher than open_file_attempts.


jrossi:
#442

jrossi:
Anyone else testing this? @ossec

jrossi:
I have not be able to test this.  Cannot someone else from @ossec look into this one?

ddpbsd:
There's a report of logs not being sent to the central manager after midnight when using this patch. I haven't confirmed if logs are sent prior to midnight though.

ddpbsd:
My crappy test setup is having a bunch of issues I'm going to blame on this. Closing for now, and will open a new pull request if I come up with anything.

PR #541, opened 2015-02-06
Update rule 31533 to fix #461

### LINKED ISSUE #461: Rule 31533 might need an update
Hi,

I'm running ossec 2.8.1 on multiple servers and usually get trouble with rule 31533 on multiple software that uses a lot of ajax POST commands (drupal, zarafa, etc).

The original rule says <rule id="31533" level="10" timeframe="20" frequency="6"> which as of my understandings means no more than 12 POST requests in 20 seconds.
As with modern software ajax is more and more used widely, shouldn't frequency be set to a much higher value like 50 ? (I actually tried 20 with Zarafa Webmail, still get that rule fired).

Regards,
Ozy.

deajan:


ChristianBeer:
But wouldn't this change render rule 31533 useless? This will allow up to 50 requests within 20 seconds. What if the attack is slower than this? Shouldn't we exclude the false positives by a common request or so (like in the rule above)?

Btw: I recently had the same problem as in #461 with a pydio installation. May first reaction was also to increase the frequency but I don't find this ideal.

deajan:
Well slow attacks are the hardest to deal with (remember slowloris ?).
I think most people have to increase maximum allowed requests anyway for this rule for their recent ajax apps to work. At least, it will work out of the box.

Do you know a better way to deal with this ?

ChristianBeer:
Wouldn't it be better to deal with those AJAX applications locally be setting up rules that ignore 31533 for those special paths? OSSEC should provide a good security out of the box. But changing this rule for all the web applications on the monitored server is weakening the security.

So I would propose to add a notice to this rule that gives a sample rule someone can put into local_rules.xml to suppress 31533 for specific paths that are known to produce a lot of requests. This way the admin deliberately makes a security decision that will not affect all web applications.

deajan:
Well the main drawback is that rule 31533 will be totally inactive for those specific paths.
What's best considering that there aren't many non AJAX apps out there today ?

ChristianBeer:
I still think it's better to have a strict default rule that can be overridden by a more flexible local rule depending on the needs of the installation.

Maybe this new local rule is basically a copy of 31533 with a more specific path and a more relaxed frequency setting. This way the sort of attack that 31533 would catch is still caught by the local copy.

I know of a lot of non AJAX websites that still need the 31533 rule as it is. This rule can for example help preventing spam profiles on phpBB message boards.

jrossi:
Lots and lots of good chat, I am leaning towards leaving this and leaving this as a local tuning.  While this change is not wrong I think it will hurt users that are not expecting it.

ChristianBeer:
@deajan Can you send me some log samples of such ajax calls for your application? I have a pydio installation that I need to fine tune rule 31533 for. Maybe I can come up with an example rule that will fit us both.

deajan:
@ChristianBeer Well i have raised the value step by step until i got my apps (Zarafa + Drupal with CKeditor / IMCE) working.
I could reactivate verbose logs in httpd and make some tests but i don't think it's necessary for me.
What values have you set rule 31533 to ? If our values match, i could create a PR with two local rules types that replace rule 31533 or enhance it with <url></url>.
The original rule 31533 could have the following text appended: 
`If this rule is triggered by an AJAX application, consider enabling rule 100002 or 100003 in local_rules.xml`

ddpbsd:
@deajan You can create a FAQ entry (http://ossec-docs.readthedocs.org/en/latest/faq/alerts.html) on ossec-docs, or even a recipe (http://ossec-docs.readthedocs.org/en/latest/cookbooks/index.html)

PR #547, opened 2015-02-13
Feature/nfs exclusion v2

reyjrar:
This implements the <skip_nfs>yes</skip_nfs> option in rootcheck and syscheck.  Since it's available, it also warns if a user attempts to iNotify on an NFS partition as that is not supported by iNotify.


jrossi:
@reyjar could you add some documentation for this?  With docs I feel people will get lost with this feature.

reyjrar:
Added documentation on : https://github.com/ossec/ossec-docs/pull/94

PR #570, opened 2015-04-06
Server side configuration to mark agents as disconnected

Pantera1989m:
I believe it would be useful if the server can be configured to mark agents disconnected after x time. In conjunction with notify_time and time_reconnect on the agent side this can be used for uptime monitoring.


awiddersheim:
It looks like something happened to this pull request and a whole of other stuff was brought in. Please scope this back down to the specific change you want to make and re-open.

PR #572, opened 2015-04-07
Better Dropbear events detection

kravietz:
This MR introduces improved detection of Dropbear events, which is especially useful if you're collecting logs from OpenWRT routers.


ddpbsd:
Without testing, i like the additions. Could you add a dropbear.ini in contrib/ossec-testing/tests? This is something we're trying to add as time permits, andadding it alongside new rules has been helpful.

kravietz:
Just added these, mostly mimicking what I found in sshd.ini. Let me know if it works. Also added one more detection rule for Dropbear.

ddpbsd:
It'll be a few days before i can test. If you don't see anything by
wednesday please ping me.

ChristianBeer:
Is it possible that the build system is wonky at the moment? I checked out this PR to my local system and ran the tests myself. The dropbear tests are passing but the pam.ini tests are failing because the pam decoder is not working. I checked that my local_decoder.xml is empty.

awiddersheim:
Retrying build. Lets see what happens.

awiddersheim:
Doesn't seem to have helped. Relevant lines are as follows:

https://travis-ci.org/ossec/ossec-hids/jobs/57781949#L1421-L1482

kravietz:
The `dropbear_rules.xml` file was not included in `ossec.conf` which I suppose is a sufficient reason for the rules not to work :) Just fixed in the commit above.

ddpbsd:
I know I asked you to add it, but can you remove the dropbear.ini file? I don't know why it's failing on travis, but it would be nice to get these additions in. The tests pass locally, but no one has been able to figure out what travis is doing to make them fail.

PR #577, opened 2015-04-09
Fixed deletion detection

aalberdi:
This modification creates a fix for detecting the deletion of files on each scan. This pull request is based on Gaelmullers syscheck improvements (pull request #5 ).
The second modification is for removing redundant code in run_check.c where that check is not needed as both give the same result.

### DIFF
--- a/src/syscheckd/create_db.c
+++ b/src/syscheckd/create_db.c
@@ -59,8 +59,17 @@ static int read_file(const char *file_name, int opts, OSMatch *restriction)
     if (lstat(file_name, &statbuf) < 0)
 #endif
     {
-        merror("%s: Error accessing '%s'.", ARGV0, file_name);
-        return (-1);
+        if(errno == ENOTDIR){
+		/*Deletion message sending*/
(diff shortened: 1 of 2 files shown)


src/syscheckd/create_db.c
```diff
@@ -59,8 +59,17 @@ static int read_file(const char *file_name, int opts, OSMatch *restriction)
...
 #endif
     {
-        merror("%s: Error accessing '%s'.", ARGV0, file_name);
-        return (-1);
+        if(errno == ENOTDIR){
+		/*Deletion message sending*/
+		char alert_msg[PATH_MAX+4];
+		alert_msg[PATH_MAX + 3] = '\0';
+		snprintf(alert_msg, PATH_MAX + 4, "-1 %s", file_name);
+		send_syscheck_msg(alert_msg);
```
aalberdi: This line sends the deletion message. It will only check that the files included in the list provided by the manager still exist. It is similar to real time but for single files.

PR #588, opened 2015-04-17
if_not_sid tag

lc4ntoni-centralway:
Hello,

this pull request is for the creation of the "\<if_not_sid\>" tag. It basically allows to setup exclusions for patterns matched within the log line. In the example below I want to match the "bar" pattern for a certain event, but for the same event I want to not raise an alert if the "foo" string is present as well.

\<group name="local,syslog,"\>

\<rule id="100002" level="0"\>
    \<if_sid\>5710\</if_sid\>
    \<match\>foo\</match\>
    \<description\>testfoo\</description\>
 \</rule\>
 
 \<rule id="100003" level="8"\>
    \<if_sid\>5710\</if_sid\>
    \<if_not_sid\>1000002\</if_not_sid\>
    \<match\>bar\</match\>
    \<description\>foo pattern exclusion\</description\>
 \</rule\>   

analysisd stores the matched sids used for comparison within an array which in my patch has a fixed size. I'm not sure this is the best option although the maximum size I think has a reasonable value.

### DIFF
--- a/src/analysisd/analysisd.c
+++ b/src/analysisd/analysisd.c
@@ -10,6 +10,7 @@
 /* ossec-analysisd
  * Responsible for correlation and log decoding
  */
+#define MATCHED_SID_ARRAY_SIZE 200000
 
 #ifndef ARGV0
 #define ARGV0 "ossec-analysisd"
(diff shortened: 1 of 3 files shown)


jrossi:
Truly sorry for not revviewing this till just now.  I was wondering if you would be willing to submit a pull request for the documention and maybe using the rules testing create a test case that show how this feature works correctly.  This would allow for us to build upon for any future changes and feel more comfortable that we would not break your feature.

lc4ntoni-centralway:
My apologies, I just saw your comment now. Sure, I will provide a new PR with better documentation and a test case

dcid:
I just tried it out and the code doesn't seem to do what it says it does. The rule specified in if_not_sid is never triggered as part of the flow. 

I also think that looping through all possible events would be very inefficient here. The ideal setup for a if_not, would be to create a third rule matching the pattern you want to filter out and setting it to level =0.

Declining.

PR #593, opened 2015-04-20
Rectification of user_id in prelude.c

Schnaffon:
The user id was missing in prelude alerts because user_id was misspelled.


jrossi:
Thank you - did not know anyone was still using prelude.  Was thinking about pulling it out and moving it into a zeromq external module.

Schnaffon:
You're welcome :)

We're actually currently working in the SECEF project and trying to improve IDMEF alerts and prelude modules.
A new version for libprelude will be out soon and we're working on new possibilities to improve incident detection.

Thank you for accepting these small corrections. They may be followed by others.

PR #600, opened 2015-04-27
readme: quick install upgraded to 2.8.1

gasp:


awiddersheim:
I can understand updating the version number by why change the root location from:

 https://github.com/ossec/ossec-hids/releases/download

To:
 
 https://github.com/ossec/ossec-hids/archive/

gasp:
Because the old link was pointing to a #404 error

awiddersheim:
Weird. I could have sworn I tried the old link and was able to get 2.8.1 fine which prompted me to ask my question but you are right, it is broken. Sorry.

Issue #574, opened 2015-04-07
RPM spec files


nigelhorne:
When will the spec files for the server be updated from 2.7 to 2.8?

awiddersheim:
I'm not sure anyone really maintains these SPEC files. Most people just use atomic's repository and RPMs for OSSEC:

https://www5.atomicorp.com/channels/atomic/

Pull requests are welcome if you'd like to update and maintain these other wise I'd suggest using atomic's version of the RPMs.

@ossec Maybe we should consider removing these.

dsvetlov:
Hi, all!

I was very surprised, when saw this issue. I have try to find specs in last year and it was unsuccessful. In finish I just take Atomic's src.rpm, unpack it, disable Atomic's patches and build vanilla OSSEC rpms. 

I think it'll be better to keep specs inside OSSEC project. How knows what will happen with Atomic. Anyway I'll prefer to use original specs.

I hope, I'll find time in next weekend to check what is wrong with current SPECS, and how to use it. Then I will make pull request.

atomicturtle:
@awiddersheim agreed here, we should just drop this from the package

nigelhorne:
On 4/11/15 12:07 PM, Andrew Widdersheim wrote:
>
> I'm not sure anyone really maintains these SPEC files. Most people 
> just use atomic's repository and RPMs for OSSEC:
>
> https://www5.atomicorp.com/channels/atomic/
>
> Pull requests are welcome if you'd like to update and maintain these 
> other wise I'd suggest using atomic's version of the RPMs.
>
> @ossec <https://github.com/ossec> Maybe we should consider removing these.
>

Some of the parameters of ossec (e.g. maximum number of agents) can only 
be tweaked by rebuilding, so using a pre-built RPM is not an option.

KR,

-Nigel

Issue #575, opened 2015-04-07
SPEC files don't work in a Mock build enviroment


nigelhorne:
They assume that the target area on the machine where the server is to be deployed is the same as on the machine where you're creating the RPM.

awiddersheim:
See #574.

SoftWarrior:
Great job though...can I know what
U ppl are working on?

atomicturtle:
Merging this with #574

SoftWarrior:
True

PR #606, opened 2015-05-03
New RPM spec file for issue #574

dsvetlov:
Hello!

I have make new spec file from atomicorp's one. This files uses only vanila OSSEC code from github repo and default configs from it. 

I test it briefly. Seems it works. RPM builds, server and client from it works.

Hope it will help resolve issues #574 and #575


awiddersheim:
Restarted the build. Last one failed downloading packages or something.

dsvetlov:
Hi! Is there any problems with mergin this? How can I help?

awiddersheim:
Seems okay to me. @atomicturtle any objections?

atomicturtle:
Since this is based off my WIP spec file for 2.9, I'll go over the other things left to finish up here.
1) On 2.9 we're only getting partial relro via full relro we had on 2.8
2) I hadnt finished verifying all the permissions were set correctly
3) The postgres module needs to be implemented
4) cosmetics, I was planning on renaming it from "ossec-hids-client" to "ossec-hids-agent"
5) conditional logic for the systemd service files
6) Letters in a version field (2.9.0b3). This is going to look "newer" than "2.9.0" on an upgrade event.
7) the authd certificate generation step isnt complete
8) the geoip or other conditionals are better used via "with_XXXX" instead of a binary 0/1 case. This is so it can be passed in as a build time argument with mock or rpmbuild
9) is there some reason you dropped the rules.d & decoders.d directories?

dsvetlov:
1) On 2.9 we're only getting partial relro via full relro we had on 2.8
Can you, please, explain? I don't understand. What is relro?

2) I hadnt finished verifying all the permissions were set correctly
I'm too, but I install packages, and it runs well last weeks.

3) The postgres module needs to be implemented
4) cosmetics, I was planning on renaming it from "ossec-hids-client" to "ossec-hids-agent"
Will do that.

5) conditional logic for the systemd service files.
I was planed do this, when request #619 will be merged. Install.sh also need to be edited.

6) Letters in a version field (2.9.0b3). This is going to look "newer" than "2.9.0" on an upgrade event.
7) the authd certificate generation step isnt complete
8) the geoip or other conditionals are better used via "with_XXXX" instead of a binary 0/1 case. This is so it can be passed in as a build time argument with mock or rpmbuild
Will fix that.

9) is there some reason you dropped the rules.d & decoders.d directories?
Default config of ossec do not contain rules.d & decoders.d directories. So, rpm will create this dirs, but use default config, nothing will work. I'm plan add it in next pull requests simultaneously in ossec.conf, install.sh and specs.

opoplawski:
This seems to have stalled.  Any movement here?

atomicturtle:
Yeah, its living here at the moment as part of the QA  step I'm using to tag the RC builds: http://updates.atomicorp.com/channels/source/ossec-hids/ossec-hids.spec

atomicturtle:
Out of date, we did merge various iterations of this already

Issue #603, opened 2015-04-30
sid 5402 incorrectly alerts for less than 3 failed attempts


mikedowney01:
When running sudo, if a user fails to authenticate only once or twice and exits without attempting authentication again, sid 5402 will incorrectly alert that "Successful sudo to ROOT executed". This does not alert for 3 failed attempts, due to sid 5401 matching first. Matching "incorrect password attempt" will solve this problem. I would be more than happy to submit a pull request, but I'm not sure if the better solution would be to edit rule 5401 so that it alerts for any number of incorrect password attempts, or to add another rule before 5402, changing it's sid to 5403 and potentially causing issues with local_rules.

ddpbsd:
This is definitely a tricky one. I think the easiest answer is to change 5401 to log all failed attempts, and create a rule looking for 3 instances of 5401, but I'd love to hear other opinions on this.

mikedowney01:
I think you're right. The referenced pull request will take care of the problem. I kept the old rule pretty much the same, just changed the sid and added an <if_sid> for the new 5401. I think it's important to keep this rule as is, instead of matching for three instances of 5401. The reason being that 3 failed sudo attempts will only generate one log, which this rule is matching with "3 incorrect password attempts".

PR #608, opened 2015-05-08
Correct logging of failed sudo attempts noted in issue #603

mikedowney01:
This will fix the incorrect logging of "successful login attempts" form a failed sudo attempt > 3 times.

### DIFF
--- a/contrib/ossec-testing/tests/sudo.ini
+++ b/contrib/ossec-testing/tests/sudo.ini
@@ -7,3 +7,23 @@ rule = 5403
 alert = 4
 decoder = sudo 
 
+[Failed attempt to run sudo]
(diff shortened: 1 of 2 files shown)


jrossi:
Sorry for taking so long.  Was wondering if you click added a unit test for this fix so it does not get reverted in the Future.  

https://github.com/ossec/ossec-hids/blob/master/contrib/ossec-testing/runtests.py And https://github.com/ossec/ossec-hids/blob/master/contrib/ossec-testing/tests/sshd.ini would be the places to look.  If you have any question.  Please let us know and thank you for taking the time and work on the pull request.

mikedowney01:
Not a problem. I believe this is what you were looking for. I couldn't seem to add a check for rule 5402 "Successful sudo to ROOT executed" since the test would always fail and see this as 5403 "First time user executed sudo".

mikedowney01:
Is there anything I can do to help move this along?

PR #612, opened 2015-05-15
Update ssh_asa-fwsmconfig_diff

brentmorris253:
Changed 'send "show running-config"' to 'send "more system:running-config"'.  This allows OSSEC to detect changes to preshared keys, such as RADIUS and Tunnel-Groups.  Revealing this information allows OSSEC to do a more thorough job of inspecting the running configuration for changes.  Previously, this information was masked behind five asterisks.

Reversed the order of the output and inserted two sleep commands.  This eliminated false positive alerts due to an inconsistent number of line feeds at the end of the file.  Adding two sleep commands and moving the larger portion of the data to the top of the file has eliminated the issue in my testing.

### DIFF
--- a/src/agentlessd/scripts/ssh_asa-fwsmconfig_diff
+++ b/src/agentlessd/scripts/ssh_asa-fwsmconfig_diff
@@ -77,7 +77,7 @@ if {[string compare $pass "x"] == 0} {
 
 # SSH to the box and pass the directories to check
 if [catch {
-    spawn ssh -c des $hostname
+    spawn ssh -q -c des $hostname
 } loc_error] {
     send_user "ERROR: Opening connection: $loc_error.\n"
     exit 1;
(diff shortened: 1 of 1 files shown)


mstarks01:
Would that not also mean that the keys were potentially exposed in an email alert?

brentmorris253:
mstarks01 - yes.  Good point.  It seems my quest for a deeper look at the cisco configuration is flawed as changed pre-shared-keys may be included in the email alert.  Should I nix this pull request?  The only other way to do this and comply with some security policies would be to write a flag into ossec.conf to enable/disable including changes in alerts.  bummer!

mstarks01:
I think a configuration flag is a good idea. Your idea has merit, but it unfortunately creates a vulnerability in the default configuration. OSSEC should act securely out of the box, methinks.

wayridd:
Please remove my email address. I did not sign up for your emails, remove my name ASAP, thank you

jrossi:
I don't think we should do something like this on security device.  Going to close for please re-open if we will want do something wiht this PUll Request.

PR #615, opened 2015-05-16
Binding src IP to 'local_ip' config value in agentd.

AndrewKant:
In mulihomed host environment we have a big problem with binding agent to correct ip. By default agentd used ip-addr of interface, from which sented ip-packets. If we have tunnel or dynamic ip, server will be reject such packets.

I prepared and tested little patch for solve this problem by use 'local_ip' config value:
```
<ossec_config>
  <client>
    <server-ip>10.0.0.82</server-ip>
    <local_ip>192.168.0.101</local_ip>
  </client>
....
```

### DIFF
--- a/src/client-agent/start_agent.c
+++ b/src/client-agent/start_agent.c
@@ -84,10 +84,13 @@ int connect_server(int initial_id)
         /* IPv6 address */
         if (strchr(tmp_str, ':') != NULL) {
             verbose("%s: INFO: Using IPv6 (%s).", ARGV0, tmp_str);
-            agt->sock = OS_ConnectUDP(agt->port, tmp_str, 1);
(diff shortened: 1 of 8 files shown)


jacobslife:
I second this pull request. In any dynamic routing environment the inability to specify a source address for agents becomes a big issue.

GitHubEmail:
I third this request! I chased this issue around for a while myself. This is much needed functionality for reliable OSSEC server <-> agent communications. Without it, packets may never reach their target.

Namsep:
Don't forget about IPv6

aquerubin:
I've added this functionality to the IPv6 fork/pull-request.

aquerubin:
Temporarilly backing out the change to the IPv6 fork as it broke maild.

Issue #609, opened 2015-05-13
lock functions in firewall-drop.sh and host-deny.sh cause severe load spike when stopping ossec


aquerubin:
On Linux agents with active response enabled and the agent is subject to
numerous probes resulting in hundreds of accumulated addresses added to both
the hosts.deny file and iptables over a long period, when ossec is stopped
the system load skyrockets due to the near-simultaneous locking done by
numerous host-deny.sh and firewall-drop.sh processes removing the previously
added addresses.  In extreme cases, the system becomes unresponsive for tens
of minutes at a time while ossec attempts to remove the addresses.

I am currently testing whether sleeping on every iteration of the lock loop would improve things.

aquerubin:
It turns out sleeping on every iteration of the lock loop isn't enough.  A backoff mechanism is needed.

PR #624, opened 2015-06-14
Wait between active response flushes during shutdown to prevent the processes from entering sleep state right away.

aquerubin:
This further addresses issue #609 by preventing the processes from starting up simultaneously and running concurrently in the first place. While the previous patch addressed system load between processes that were already running simultaneously, this fix avoids that altogether.

### DIFF
--- a/src/os_execd/execd.c
+++ b/src/os_execd/execd.c
@@ -49,6 +49,10 @@ static void help_execd()
 /* Shut down execd properly */
 static void execd_shutdown(int sig)
 {
+    struct timespec delay;
+    delay.tv_sec = 0;
+    delay.tv_nsec = AR_FLUSH_INTERVAL;
+
     /* Remove pending active responses */
     merror(EXEC_SHUTDOWN, ARGV0);
 
@@ -63,6 +67,10 @@ static void execd_shutdown(int sig)
         /* Delete current node - already sets the pointer to next */
         OSList_DeleteCurrentlyNode(timeout_list);
         timeout_node = OSList_GetCurrentlyNode(timeout_list);
+
+        /* Delay to prevent system from being overwhelmed by locked
+           processes if there are many nodes */
+        nanosleep(&delay, NULL);
     }
 
     HandleSIG(sig);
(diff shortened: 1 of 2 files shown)


jrossi:
Sorry about being so slow with updates.   I have often thought about doing this but was worried the response time and queue would back under load.  I worry that this little change would have a huge effect on some people like @atomicturtle.  Is this a real issue? @ossec crew any input? 


I had started something around batching and stuff a while ago but needed more eyes and love and time. https://github.com/jrossi/ossec-hids/tree/ossec-execd-lua-scripts-state?files=1

awiddersheim:
I'm also worried. Would be nice if someone with a good amount of throughput could test.

aquerubin:
On Fri, 3 Jul 2015, Andrew Widdersheim wrote:

> I'm also worried. Would be nice if someone with a good amount of 
> throughput could test.

Amount of processing throughput is not the real question.  The desired end 
state is that all AR records are flushed from the system as quickly and 
painlessly as possible during OSSEC shutdown.  The question is really how 
effectively to use that processing throughput to achieve that desired 
state.  Is it better to instantly startup a whole bunch of processes that 
are almost certain to go into lock/sleep states of varying duration and 
walk away (letting the system sort things out), or, simply run the 
processes sequentially as fast as each flush can be completed (completely 
avoiding lock/sleep cycles altogether).  Empirical testing when the number 
of AR entries number in the hundreds shows the former is a really bad 
idea.  But don't take my word for it, do your own rigorous load testing 
during shutdown and verify for yourself what the effects in either case.

Remember this patch just adds a wait() only during OSSEC shutdown.  It's a 
bit ironic that if you had a situation where a server is undergoing a 
persistent attack that creates lots of AR entries, certain actions that 
might be taken to quiesce the server and shutdown OSSEC as part of that 
action actually may make the situation worse by causing a self-induced 
DoS.


Antonio Querubin
e-mail:  tony@lavanauts.org

jrossi:
@aquerubin you are correct and I was and am completely WRONG :) let me start again ;)

awiddersheim:
Yeah I totally misunderstood this as well.

PR #633, opened 2015-07-06
Compile with hardening flags

bchurchill:
These flags mitigate the risk of zero-day memory corruption vulnerabilities, like buffer overflow attacks.  In brief, the options added are:

-fpic -pie
Compiles position independent code, to take advantage of ASLR.  This makes it harder to write exploits because it requires an information leak bug or brute force attack in addition to the memory corruption exploit.

-fstack-protector-all --param ssp-buffer-size=4
Ensures a canary is placed on the stack any time there's a potentially vulnerable function.  This mitigates many stack-based overflows.

-Wl,-z,relro,-z,now
Makes relocation tables read-only, making it harder to do a return-to-libc style attack by rewriting a relocation entry.

-Wformat -Wformat-security
Complains loudly if there's an obvious format string injection vulnerability.

-D_FORTIFY_SOURCE=2
Adds some extra checks on buffer lengths.

You can read about some of these options in section 5 of https://wiki.debian.org/Hardening.  Some details are also at https://wiki.ubuntu.com/ToolChain/CompilerFlags?action=show&redirect=CompilerFlags.


bchurchill:
I can't reproduce some of these environments locally, so I may need to submit a few more pull requests to get the kinks worked out.

atomicturtle:
For what its worth we've been building the rpms with this for many years now, and a good test suite to look at the settings is: http://www.trapkit.de/tools/checksec.html

As it stands right now, we still need to make some changes to the source tree to get this working right in 2.9 so I'm going to re-open this one.

bchurchill:
At first I didn't realize I could make a commit and have travis rebuild the pull request, so I thought I'd have to submit others (which is why I closed this one earlier).  Now that the build is working, I think it's ready for someone else to review.

ddpbsd:
I'm seeing a lot of messages I'd rather not see when compiling with this diff:
```
gcc: -z: linker input file unused because linking not done
gcc: relro: linker input file unused because linking not done
gcc: -z: linker input file unused because linking not done
gcc: now: linker input file unused because linking not done
```

bchurchill:
Ok, I'll look into that.

atomicturtle:
Looks like this doesnt get ossec-dbd either:

       ossec-dbd  27528 No RELRO          Canary found           PaX enabled   No PIE

atomicturtle:
Oh and authd is only partial:
     ossec-authd  29936 Partial RELRO     Canary found           PaX enabled   PIE enabled

sempervictus:
Is this still live, or have the v3+ compiler options included the relevant hardening bits in other commits?
Separately, I'd probably suggest buffer overflow and struct randomization under GCC, as well as building with selfrando for fine grained randomization per run so long as mprotect is not in play (maybe using mmap for those cases). I imagine clang builds could benefit from CFI and other hardening approaches as well.

ddpbsd:
I don't think they're on by default, but they can be set during compile time. @atomicturtle might be using some of these for the packages. I imagine some of these can be set for some systems, but whether AIX and similar systems would work is a mystery.

Issue #301, opened 2014-09-16
OSSEC agent on  64-bit Windows


bvb09michel:
There is no 64-bit agent for windows! 

The 32-bit agent is not able to access the important windows\system32 directory. The same problem exist for access to some important registry hives. 
In the ossec.conf file many are active by default of these important entries. 
The reason (32-bit application on 64-bit OS) is described here: https://groups.google.com/forum/#!topic/ossec-list/E7YgAoVcFkU 

You can show the problem with the telnet.exe. This file only exists in the windows\system32 directory (64 bit exe) and not in the windows\syswow64 directory (32 bit exe) when the Telnet feature is installed. 
The telnet.exe file ist configured in the ossec.conf. 
Thus appears permanently in the ossec.log the error: 
2014/09/08 03:33:18 ossec-agent: WARN: Error opening directory: 'C:\Windows/System32/telnet.exe': No such file or directory 
But in reality it exists, of course. 

So the problem is, that are not monitored in the windows\system32 directory in reality. 
If the monitored file is defined in the ossec.conf, this is monitored in reality in the windows\syswow64 directory. 
The same problem applies to the Registry!

123BLiN:
I use next workaround for x64 registry if this could help someone (I spent two days to find it :( ):
In ossec.conf on client:
   <localfile>
    <log_format>full_command</log_format>
    <command>reg query "HKLM\SOFTWARE\my software" /s /f "Log Level" /reg:64</command>
  </localfile>
or simple /s for all subkeys
on server in local_rules.xml:
<rule id="100007" level="8">
    <if_sid>530</if_sid>
    <match>ossec: output: 'reg query "HKLM\SOFTWARE\my software" /s /f "Log Level"</match>
    <check_diff />
    <description>My software Log Level settings changed</description>
</rule>
/reg:64 setting is not documented in Microsoft and I dont remeber where did I find it, it make reg.exe to use x64 Registry view.

bvb09michel:
Hi 123BLiN, 
thanks for the workaround. 

greetings Michael 




Von:    123BLiN <notifications@github.com>
An:     ossec/ossec-hids <ossec-hids@noreply.github.com>, 
Kopie:  bvb09michel <michael.lange@fiducia.de>
Datum:  29.04.2015 00:20
Betreff:        Re: [ossec-hids] OSSEC agent on  64-bit Windows (#301)



I use next workaround for x64 registry if this could help someone (I spent 
two days to find it :( ):
[...]

PR #639, opened 2015-07-16
Small fix to enable monitoring of the Windows 64bit registry view

secnerd:
Hi,

firstly let me apologise - I'm not a coder and I've never used git before so please be patient :)

The attached changes modifies the calls to RegOpenKeyEx. The code is effecively duplicated to allow modifications to the samDesired parameter - the first call uses the KEY_WOW64_64KEY access mask (bitwise OR'ed with the existing KEY_READ mask) to request access to the 64bit registry view the second specifically requests access to the 32bit registry view using the KEY_WOW63_32KEY access mask. Both are ignored on 32 bit systems which will probably cause some duplication. Maybe someone with slightly more clue than I can see an easy way to get round the issue... or maybe point me in the direction of how to address that issue.

Also it's worth noting that in order to get the windows agent to build (./gen_win.sh && cd ../win-pkg/ && ./make.sh) I had to comment out lines 347 to 360 in win-pkg/seechange.c - these lines call symlink functions which aren't supported on Windoze.

Hope that helps in some way. The 64bit registry monitoring is a huge issue for me, we have almost no 32bit Windows left in our environment and plenty of sensitive 64bit systems!

One further point - this goes some way towards addressing the issues in #301 however I don't believe that simply creating a 64 bit windows agent would fix all the issues as a 64bit agent with unmodified calls to *RegOpenKeyEx* would only operate on the 64bit registry view and ignoring the 32bit view.

Secnerd

p.s. Please ignore the pull request against the v2.8.2 branch - had some issues with my build environment (32bit mingw not 64bit!)

### DIFF
--- a/src/error_messages/error_messages.h
+++ b/src/error_messages/error_messages.h
@@ -111,7 +111,8 @@
 #define SK_INV_MSG      "%s(1755): ERROR: Invalid syscheck message received."
 #define SK_DUP          "%s(1756): ERROR: Duplicated directory given: '%s'."
 #define SK_INV_REG      "%s(1757): ERROR: Invalid syscheck registry entry: '%s'."
-#define SK_REG_OPEN     "%s(1758): ERROR: Unable to open registry key: '%s'."
+#define SK_REG_OPEN     "%s(1758): ERROR: Unable to open registry key using 32 bit registry: '%s'."
+#define SK_REG_OPEN64   "%s(1759): ERROR: Unable to open registry key using 64 bit registry: '%s'."
 
 /* analysisd */
 #define FTS_LIST_ERROR   "%s(1260): ERROR: Error initiating FTS list"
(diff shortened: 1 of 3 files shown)


dalini:
any plans to merge this in?

ddpbsd:
I don't really have a way to test it, so sure. Let's merge it and see what happens.

mkgaao:
Hi, I had these changes applied to v2.8.3 and running on Windows 2008 R2 64 bit system. Most of the registry errors has gone, however few registry entry were erroring. I am going to have the agent running overnight and recheck.

ossec-agent(1758): ERROR: Unable to open registry key using 32 bit registry: 'Software\Microsoft\Windows NT\CurrentVersion\Windows\AppInit_DLLs'.
ossec-agent(1758): ERROR: Unable to open registry key using 32 bit registry: 'Software\Microsoft\Windows NT\CurrentVersion\Winlogon\AutoAdminLogon'.
ossec-agent(1758): ERROR: Unable to open registry key using 32 bit registry: 'Software\Microsoft\Windows NT\CurrentVersion\Winlogon\DefaultPassword'.

PR #651, opened 2015-08-12
Use the full version number.

ddpbsd:
I'm not aware of a reason we were only using 2.8 instead of 2.8.2, but there have been requests for the full version number (issue 644).


dmccarney:
I'm not sure if the oversight was intentional or not but it appears the value used in [src/shared/help.c's `print_version() function`](https://github.com/ossec/ossec-hids/blob/eb69de7918e33ae7d806a5a14c7d5ca2f0947ad5/src/shared/help.c) was overlooked. 

I think the `__version` define in [src/headers/defs.h](https://github.com/ossec/ossec-hids/blob/master/src/headers/defs.h#L39) also needs to be updated to `2.8.2`.

I found this PR and the accompanying issue after realizing that `-V` wasn't an effective way to learn the version of the OSSEC services post-install. Being able to find the full version using `-V` is a use-case I care about seeing supported. Thanks for the PR.

ddpbsd:
@dmccarney Thanks for pointing that out. I've updated the defs.h definition.

dmccarney:
@ddpbsd :+1:

PR #659, opened 2015-08-24
Same dev

doke2:
Add directories option to keep scan on same device, ie mountpoint.
Similar to -mount flag to unix find(1) command.


doke2:
This should let you scan root (/) but automatically avoid going into /proc, /mnt/cdrom, nfs mounts, etc.

PR #661, opened 2015-08-25
Ignore glob

doke2:
Allow glob patterns in ignore tags.

    <ignore type="glob">/var/log/www/*.log</ignore>

Also displays monitoring options and ignore lists better on startup.

    2015/08/25 16:00:07 ossec-syscheckd: INFO: Monitoring directory: '/etc', with options perm | size | owner | group | md5sum | sha1sum.
    2015/08/25 16:00:07 ossec-syscheckd: INFO: ignoring: '/etc/mtab'
    2015/08/25 16:00:07 ossec-syscheckd: INFO: ignoring glob: '*.log'
    2015/08/25 16:15:37 ossec-syscheckd: read_file ignoring '/var/log/yum.log' for glob '*.log'


dcid:
Can't we use the sregex option to specify: /var/log/www/\S+.log? Already there in the code and achieves the same...

dcid:
We already have the sregex that does the same thing. Not merging.

PR #662, opened 2015-09-07
Don't consider an empty client.keys to be a failure condition

chewi:
client.keys is already reloaded each time a given key is not found in memory so there's no harm in this file being empty. In fact, it's downright annoying if you're using authd because you have to wait for the first agent to register and then manually restart the server before they can start communicating. Removing this check would make the Chef cookbook less clunky.

Disclaimer: I haven't tested this at all because I've already sunk too much time into the cookbook. The change seems simple enough though.

[Review on Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/662)

### DIFF
--- a/src/os_auth/main-client.c
+++ b/src/os_auth/main-client.c
@@ -343,8 +343,7 @@ int main(int argc, char **argv)
                         FILE *fp;
                         fp = fopen(KEYSFILE_PATH, "w");
                         if (!fp) {
-                            printf("ERROR: Unable to open key file: %s", KEYSFILE_PATH);
-                            exit(1);
+                            ErrorExit(FOPEN_ERROR, ARGV0, KEYSFILE_PATH, errno, strerror(errno));
                         }
                         fprintf(fp, "%s\n", key);
                         fclose(fp);
--- a/src/os_auth/main-server.c
+++ b/src/os_auth/main-server.c
@@ -306,8 +306,7 @@ int main(int argc, char **argv)
 
     fp = fopen(KEYSFILE_PATH, "a");
     if (!fp) {
-        merror("%s: ERROR: Unable to open %s (key file)", ARGV0, KEYSFILE_PATH);
-        exit(1);
+        ErrorExit(FOPEN_ERROR, ARGV0, KEYSFILE_PATH, errno, strerror(errno));
     }
     fclose(fp);
 
--- a/src/os_crypto/shared/keys.c
+++ b/src/os_crypto/shared/keys.c
@@ -13,9 +13,20 @@
 #include "os_crypto/blowfish/bf_op.h"
 
 /* Prototypes */
+static void __realloc_keys(keystore *keys) __attribute((nonnull));
 static void __memclear(char *id, char *name, char *ip, char *key, size_t size) __attribute((nonnull));
 static void __chash(keystore *keys, const char *id, const char *name, char *ip, const char *key) __attribute((nonnull));
 
+
+static void __realloc_keys(keystore *keys)
+{
+    /* Allocate for the whole structure */
+    keys->keyentries = (keyentry **)realloc(keys->keyentries,
+                                            (keys->keysize + 2) * sizeof(keyentry *));
+    if (!keys->keyentries) {
+        ErrorExit(MEM_ERROR, __local_name, errno, strerror(errno));
+    }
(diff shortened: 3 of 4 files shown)


ddpbsd:
Can someone test this? I think it's a great change, but want to make sure it works first. (adding it to my list of things to do when I can make time)

reyjrar:
I think this may break some stuff.  This function is called by the agent and the server.  I agree, on the server, this shouldn't be fatal, but I'm not so sure on the agent if this would fly.

chewi:
I thought about sticking `ifdef CLIENT` around it but I see that agentd is built for `TARGET=server` as well. It would somehow need to determine whether it is an agent or server process at runtime.

chewi:
It's not fantastic but since `ARGV0` is hardcoded, how about this?

```c
if (strcmp(__local_name, "ossec-remoted") != 0) {
  …
}
```

chewi:
Rebased.

chewi:
I finally gave this a try myself and found it segfaulted because `keys->keyentries` wasn't initialised. I've reworked it and now ossec-remoted at least doesn't crash any more.

chewi:
I've now also got it to create client.keys on startup like authd does. This has to happen before dropping privileges because ossecr should not have write access. The file permissions could still be tightened up (currently root:root and 644) but authd doesn't make any effort to do this and the directory permissions at least prevent world read access. One for later perhaps.

chewi:
Any chance of getting this merged now? I haven't seen any ill effects since and yet another person asked about this problem on the mailing list.

dcid:
It looks good now, except for this function:

__realloc

Really confusing to use a standard C function name for this. Can you change to something more adequate to what it does? (allocate key mem)

chewi:
Fair enough, renamed it to __realloc_keys.

jrossi:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/os_crypto/shared/keys.c, line 262 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/662#-KCfqszlIOST094H_TcC:-KCfqszlIOST094H_TcD:2021371641) ([raw file](https://github.com/ossec/ossec-hids/blob/470de317250bbb2a0d62b8ba5015d897f9924b70/src/os_crypto/shared/keys.c#L262)):*
I dislike this for some reason.  This does not mean it's wrong just something feels wrong that this is needed.  

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ossec/ossec-hids/662)*

jrossi:
Sorry i am taking so long need to review some more things within the code base. 

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ossec/ossec-hids/662#-:-KCftqtsmu2sMa_iGJcj:1248948002)*

chewi:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/os_crypto/shared/keys.c, line 262 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/662#-KCfqszlIOST094H_TcC:-KDwhdVKlL2SGo_K5zlG:1323326475) ([raw file](https://github.com/ossec/ossec-hids/blob/470de317250bbb2a0d62b8ba5015d897f9924b70/src/os_crypto/shared/keys.c#L262)):*
Believe me, I tried to think of a better way but couldn't find one. You're welcome to try.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ossec/ossec-hids/662)*

chewi:
I'm concerned that 2.9 may get released without this and that would be very frustrating. The code smell mentioned above isn't a showstopper and could always be improved later.

atomicturtle:
Thanks @chewi — this was superseded by remoted.pass_empty_keyfile /
OS_PassEmptyKeyfile() (see PR #1628 and authd's OS_PassEmptyKeyfile() call).
Empty client.keys is allowed when that flag is set; agents still fail closed.
To get the authd/Chef behavior without a remoted restart, set in
internal_options.conf:

  remoted.pass_empty_keyfile=1

PR #663, opened 2015-09-11
fix debug message before parsing agent.conf to give the right filename

doke2:


ddpbsd:
What file is listed if there is no agent.conf on that agent?

doke2:
All this does is correct one minor debug message.  Without this patch,
it will list the normal config file (twice).  With this patch, it will list
the AGENTCONFIG file, as defined in headers/defs.h.  Either way, agents
will always try to read the AGENTCONFIG file.  If there is no agent.conf,
the read will warn but not fail, and the program will continue.  Here's the
defs.h line, and a diff with expanded context.


    #define AGENTCONFIG     DEFAULTDIR "/etc/shared/agent.conf"


    diff --git a/src/syscheckd/config.c b/src/syscheckd/config.c
    index 6e9092a..1d1a6f7 100644
    --- a/src/syscheckd/config.c
    +++ b/src/syscheckd/config.c
    @@ -41,21 +41,21 @@ int Read_Syscheck_Config(const char *cfgfile)
	 syscheck.prefilter_cmd  = NULL;
     
	 debug2("%s: Reading Configuration [%s]", "syscheckd", cfgfile);
     
	 /* Read config */
	 if (ReadConfig(modules, cfgfile, &syscheck, NULL) < 0) {
	     return (OS_INVALID);
	 }
     
     #ifdef CLIENT
    -    debug2("%s: Reading Client Configuration [%s]", "syscheckd", cfgfile);
    +    debug2("%s: Reading Client Configuration [%s]", "syscheckd", AGENTCONFIG);
     
	 /* Read shared config */
	 modules |= CAGENT_CONFIG;
	 ReadConfig(modules, AGENTCONFIG, &syscheck, NULL);
     #endif
     
     #ifndef WIN32
	 /* We must have at least one directory to check */
	 if (!syscheck.dir || syscheck.dir[0] == NULL) {
	     return (1);


Dan Parriott <notifications@github.com> said:
>What file is listed if there is no agent.conf on that agent?

                                               Cheers,
                                               Doke

ddpbsd:
Cool, as long as it doesn't mention the agent.conf when it reads the ossec.conf I'm ok with the change.

PR #668, opened 2015-09-21
Fix truncated "action" on certain Cisco PIX/ASA logs

brentmorris253:
This fixes the Cisco decoder from truncating the action.  Prior to this change, the decoder only picks up the first letter of the word 'denied'.

This can be verified with the existing sample logs - %PIX-3-710003: TCP access denied by ACL from 216.39.220.130/54065 to outside:62.192.113.98/ssh


brentmorris253:
Sorry, I am new to github.  This should be two pull requests.  One fixing the pix rules and one fixing the ms-se_rules.xml

PR #676, opened 2015-10-05
PHP rule checks on apache 2.4 logs

dopefish:
Currently the rules for grouping php warnings/errors are only run on apache 2.2 and nginx logs, this change adds apache 2.4 logs to the list.


atomicturtle:
Looking good here. Thanks for the contribution!

PR #685, opened 2015-11-09
remove the spaces of the equal sign (problem with debian)

hregis:


ddpbsd:
2.8.3 is already released. This is fixed in the main line, and the possible 2.9 branch. I'm not planning on a 2.8.4, but even if someone was, I don't think this belongs in the 2.8.3 branch. Also, I wish someone had brought this up when 2.8.3 was being prepped. I would have gladly included the fix. I don't use this, and it's been fixed so long in master that I totally forgot about it.

hregis:
ok thanks

PR #689, opened 2015-11-13
Allow maild to send through a sendmail-like executable

chewi:
Users have long called for TLS support when sending mail from OSSEC. This is not trivial to add directly but this commit enables that support by delegating the task to a sendmail-like executable such as SSMTP or Msmtp.

maild detects this mode of operation by checking for a / at the start of the smtp_server setting. It uses popen to launch the executable and passes the raw message to it via stdin. sendmail's -t argument tells
it to parse the headers given in the message so smtp_server will typically be set to something like `/usr/sbin/sendmail -t`.

In order for this to work, maild needs to be started without chrooting as it traditionally has done. It will therefore no longer chroot if it detects a / at the start of smtp_server.

It is best to view this commit with `--ignore-all-space` as there are far few changes than the regular diff would imply.

### DIFF
--- a/src/agentlessd/main.c
+++ b/src/agentlessd/main.c
@@ -146,7 +146,8 @@ int main(int argc, char **argv)
         ErrorExit(SETUID_ERROR, ARGV0, user, errno, strerror(errno));
     }
 
-    debug1(PRIVSEP_MSG, ARGV0, dir, user);
+    debug1(CHROOT_MSG, ARGV0, dir);
+    debug1(PRIVSEP_MSG, ARGV0, user);
 
     /* Signal manipulation */
     StartSIG(ARGV0);
--- a/src/analysisd/analysisd.c
+++ b/src/analysisd/analysisd.c
@@ -434,7 +434,8 @@ int main_analysisd(int argc, char **argv)
     }
 
     /* Verbose message */
-    debug1(PRIVSEP_MSG, ARGV0, dir, user);
+    debug1(CHROOT_MSG, ARGV0, dir);
+    debug1(PRIVSEP_MSG, ARGV0, user);
 
     /* Signal manipulation */
     StartSIG(ARGV0);
--- a/src/config/global-config.c
+++ b/src/config/global-config.c
@@ -441,10 +441,14 @@ int Read_Global(XML_NODE node, void *configp, void *mailp)
         } else if (strcmp(node[i]->element, xml_smtpserver) == 0) {
 #ifndef WIN32
             if (Mail && (Mail->mn)) {
-                Mail->smtpserver = OS_GetHost(node[i]->content, 5);
-                if (!Mail->smtpserver) {
-                    merror(INVALID_SMTP, __local_name, node[i]->content);
-                    return (OS_INVALID);
+                if (node[i]->content[0] == '/') {
+                    os_strdup(node[i]->content, Mail->smtpserver);
+                } else {
+                    Mail->smtpserver = OS_GetHost(node[i]->content, 5);
+                    if (!Mail->smtpserver) {
+                        merror(INVALID_SMTP, __local_name, node[i]->content);
+                        return (OS_INVALID);
+                    }
                 }
                 free(Mail->smtpserver);
                 os_strdup(node[i]->content, Mail->smtpserver);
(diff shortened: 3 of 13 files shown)


chewi:
Here's a quick link to [the second commit without whitespace changes](https://github.com/ossec/ossec-hids/commit/b775b7200a8425550f43288758ef658f798b3efd?w=1).

chewi:
Rebased with minor simplification.

chewi:
Hold up, spotted an issue. I forgot to delete a symlink I initially placed to work around the chrooting. Small adjustment needed.

chewi:
Fixed! :smile:

aquerubin:
The third and fourth parameters to OS_ConnectTCP() should be dropped.

chewi:
@aquerubin, I haven't added or changed any `OS_ConnectTCP()` lines so that is a separate concern.

aquerubin:
On Wed, 2 Dec 2015, James Le Cuirot wrote:

> @aquerubin, I haven't added or changed any `OS_ConnectTCP()` lines so 
> that is a separate concern.

Understood.  I'm referring to the merge conflicts with the current 
master source.

Antonio Querubin
e-mail:  tony@lavanauts.org

chewi:
Okay, I see. I've rebased it now then.

rakesh-eltropy:
I tried to use msmtp in ossec.conf. If I use msmtp from command line its working fine from me but mail are not being sent from ossec at all.

`/usr/bin/msmtp -f "admin@.....com" -t`

```
sudo /var/ossec/bin/ossec-control status
ossec-monitord is running...
ossec-logcollector is running...
ossec-syscheckd is running...
ossec-analysisd is running...
ossec-maild is running...
ossec-execd is running...
```
ossec is running fine as well. Am I missing anything ?

chewi:
Are you using 2.9.0? It's not in earlier versions. We're still using this with ssmtp but we're still using our custom 2.8.3 build rather than 2.9.0.

Are you setting smtp_server to `/usr/bin/msmtp -t`? Don't put -f here, you can set that using email_from.

Some distros install these binaries with SGID but some don't, requiring you to be in a particular group so check the permissions.

rakesh-eltropy:
I am using 2.9.0 only and trying to run ossec as root user in Ubuntu 16.04 linux server.

chewi:
You may start it as root but it'll drop privileges to other users. I think `ossecm` is the user for this process.

rakesh-eltropy:
Could you please let me know which permission `ossecm` would require ?

chewi:
That's not what SGID means. Check with `ls -l` whether msmtp has the SGID bit set. It may look something like this. The `s` represents SGID.

```
-rwx--s--x 1 root mail 39552 May 18  2016 /usr/bin/msmtp
```

If it is set then you shouldn't need to do anything. If it's not set then you should add the `ossecm` user to the group, which is `mail` in the above example.

rakesh-eltropy:
Thanks @chewi. It was permission issue only. Its working fine now.

rodrigojilar:
Hi!
I'm currently using OSSEC v2.9.3 and I am trying to use msmtp too. I placed the path in smtp_server parameter, gave group permissions to ossecm and still can't make it work. Is there a way to debug this issue? Mail.log Is empty.

chewi:
I can't remember where ossecm logs to off the top of my head, if anywhere. Check the syslog and main OSSEC log. Also check that your configuration would cause it to mail anything at all.

rodrigojilar:
Here is my ossec.conf fragment:
<global>
    <email_notification>yes</email_notification>
    <email_to>xxxx@gmail.com</email_to>
    <smtp_server>/usr/bin/msmtp -t</smtp_server>
    <email_from>admin@xxxx.xxx</email_from>
  </global>
And it is supposed to send emails of every event.
<alerts>
    <log_alert_level>1</log_alert_level>
    <email_alert_level>1</email_alert_level>
 </alerts>
Groups of ossecm user 
ossec mail
Group permissions for msmtp
root:mail
syslog and ossec.log are actually empty about this problem.

Finally, I have msmtp working from the command line, and with other daemons without any problem.

rodrigojilar:
Nevermid, I replaced mSMTP with Postfix and everything is working fine now. Thanks for helping me.

PR #703, opened 2015-12-03
Agent installation. XML Error.

snaow:
Reverting commit: https://github.com/ossec/ossec-hids/commit/3dad0daecd2bddca5e256b3ad85a4d5f54d66f02

Agent installation fails throwing the following error:

`` ossec-logcollector(1226): ERROR: Error reading XML file '/var/ossec/etc/shared/agent.conf': XMLERR: File '/var/ossec/etc/shared/agent.conf' not found. (line 94).
``
`` 
ossec-syscheckd(1226): ERROR: Error reading XML file '/var/ossec/etc/shared/agent.conf': XMLERR: File '/var/ossec/etc/shared/agent.conf' not found. (line 94).
``
`` 
ossec-syscheckd(1226): ERROR: Error reading XML file '/var/ossec/etc/shared/agent.conf': XMLERR: File '/var/ossec/etc/shared/agent.conf' not found. (line 94).``
`` 
**agent.conf** is not present at initial agent installation that's why we need to prevent the **merror**.

If you are interested, the full trace would be:
- At src/rootcheck/config.c, src/logcollector/config.c and src/syscheckd/config.c **ReadConfig** function is called.
```
ReadConfig(modules, AGENTCONFIG, &syscheck, NULL);
ReadConfig(modules, AGENTCONFIG, &log_config, NULL);
ReadConfig(modules, AGENTCONFIG, &rootcheck, NULL);
```
- At src/config/config.c, inside ReadConfig, **OS_ReadXML** function is called

```
OS_ReadXML(cfgfile, &xml)
```

So ``cfgfile`` is ``AGENTCONFIG`` which value is *"shared/agent.conf*, the result is OS_ReadXML can't open the file and always stop agent initialization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ OSSEC is a full platform to monitor and control your systems. It mixes together
 all the aspects of HIDS (host-based intrusion detection), log monitoring and 
 SIM/SIEM together in a simple, powerful and open source solution.
 
-Visit our website for the latest information.  [www.ossec.net](http://www.ossec.net)
+Visit our website for the latest information.  [ossec.github.io](http://ossec.github.io)
 
 
 
@@ -14,8 +14,8 @@ Visit our website for the latest information.  [www.ossec.net](http://www.ossec.
 
 The current stable releases are available on the ossec website. 
 
-* Releases can be downloaded from: [Downloads](http://www.ossec.net/?page_id=19)
-* Release documentation is available at: [docs](http://www.ossec.net/doc/)
+* Releases can be downloaded from: [Downloads](http://ossec.github.io/downloads.html)
+* Release documentation is available at: [docs](http://ossec.github.io/docs/)
 
 ## Development ##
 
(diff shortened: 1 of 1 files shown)


ddpbsd:
I haven't seen it stop an agent from starting, is that what's happening to you?

snaow:
Yes. Just clone the repository, ./install.sh select agent installation wait
until finish and the error will show up.

ddpbsd:
I have done that, and it doesn't seem to affect the operations of the agent. I definitely don't want the error there, but I'm looking at other fixes at the moment.

snaow:
I can't test it right now but i was sure that the agent couldn't start
properly if there is no agent.conf file. Do you mean that the agent can
start up without agent.conf?

Anyways thanks for check it out, i really appreciate it

ddpbsd:
I don't think it stopped my agents from actually working. But the new PR changes it to a warning. I like having the additional information, but since agent.conf isn't required it shouldn't be more than a warning.

snaow:
Agree, thanks.

PR #705, opened 2015-12-07
Change the "can't load an agent.conf" error into a warning.

ddpbsd:
agent.conf isn't required, so it doesn't seem like an error for it
to not be there.
Prompted by @snaow and PR #703 

The config name dance is prompted by the CAVEAT in OpenBSD's basename(3) man page:
```
     Other vendor implementations of basename() may modify the contents of the
     string passed to basename(); this should be taken into account when
     writing code which calls this function if portability is desired.
```


ddpbsd:
Should this be INFO instead of WARN?

santiago-bassett:
+1 to INFO instead of WARN

ddpbsd:
Updated to INFO

ddpbsd:
Ping? Any comments, concerns, cash donations?

PR #709, opened 2015-12-09
allow multiple agent id in active_response

msongd:
This patch allows multiple agent_id in <active_response>, for example:

```
<active-response>
  <command>firewall-block</command>
  <location>defined-agent</location>
  <agent_id>001,002,003</agent_id>
  <rules_group>authentication_failed,authentication_failures</rules_group>
  <timeout>600</timeout>
  <repeated_offenders>30,60,120</repeated_offenders>
</active-response>
```

[Review on Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/709)

### DIFF
--- a/src/remoted/ar-forward.c
+++ b/src/remoted/ar-forward.c
@@ -137,17 +137,37 @@ void *AR_Forward(__attribute__((unused)) void *arg)
 
             /* Send to a pre-defined agent */
             else if (ar_location & SPECIFIC_AGENT) {
+                char *org_agent_list = NULL;
+                char *agent_list = NULL;
+                char *token = NULL;
+                int  agent_id_err = 0;
+
                 ar_location++;
 
-                agent_id = OS_IsAllowedID(&keys, ar_agent_id);
-
-                if (agent_id < 0) {
+                org_agent_list = agent_list = strdup(ar_agent_id);
+                //verbose("%s: [INFO] selected agent:=%s", ARGV0, org_agent_list);
+                if (org_agent_list == NULL) {
                     key_unlock();
                     merror(AR_NOAGENT_ERROR, ARGV0, ar_agent_id);
-                    continue;
+                    continue ;
                 }
+                while ((token = strsep(&agent_list, ",")) != NULL) {
+                    if (*token!='\0') {
+                        agent_id = OS_IsAllowedID(&keys, token);
+                        //verbose("%s: [INFO] send to agent_id=%d", ARGV0, agent_id);
 
-                send_msg((unsigned)agent_id, msg_to_send);
+                        if(agent_id < 0) {
+                            //key_unlock();
+                            merror(AR_NOAGENT_ERROR, ARGV0, token);
+                            agent_id_err = -1;
+                            continue;
+                        }
+                        else {
+                            send_msg((unsigned)agent_id, msg_to_send);
+                        }
+                    }
+                }
+                free(org_agent_list);
             }
 
             /* Lock use of keys */


src/remoted/ar-forward.c
```diff
@@ -137,17 +137,37 @@ void *AR_Forward(__attribute__((unused)) void *arg)
...
+                //verbose("%s: [INFO] selected agent:=%s", ARGV0, org_agent_list);
+                if (org_agent_list == NULL) {
                     key_unlock();
                     merror(AR_NOAGENT_ERROR, ARGV0, ar_agent_id);
-                    continue;
+                    continue ;
                 }
-
-                send_msg((unsigned)agent_id, msg_to_send);
+                while ((token = strsep(&agent_list, ",")) != NULL) {
```
ddpbsd: What I've found online might be old information, but it doesn't look like Solaris has strsep. I didn't try to check on AIX/HP-UX.

msongd: Seem Solaris does not have strsep(). Should I copy strsep() func from *bsd source to the patch?
If so, should I create another pull request?
I do not have Solaris and other Unix variants to test...

ddpbsd:
I think strtok(3) is the preferred method.

dcid:
Might be easier to simplify there. An easy:

```
tmpstr = agentid;
do
{
    if(*tmpstr == '\0' || *tmpstr == ',')
   {
       if(*tmpstr != '\0') { *tmpstr = '\0'; tmpstr++; }
       /* do whatever here */

       agentid = tmpstr;
   } 
   tmpstr++;
}while(*tmpstr != '\0');
```

That way you can go through all possible commas and still work across all systems.

jrossi:
@dcid Side not we might want to make this a util function someplace as it's reimpleted ALL over the place. 

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/remoted/ar-forward.c, line 154 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/709#-KC6z6FZyYvpZ0AoTKMD-r1-154:-KCfyc8BSsAgyI5DyL4a:-679002975) ([raw file](https://github.com/ossec/ossec-hids/blob/00c5d4da47d9677ad95705beadf4ee3d286b5373/src/remoted/ar-forward.c#L154)):*
I think @dcid method is inline with what we do in other areas of the code.  

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ossec/ossec-hids/709#-:-KCfymh7V9PKFOXlxoP1:-887945548)*

jrossi:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ossec/ossec-hids/709)*

aquerubin:
Something like this using strtok should work.  This would be useful for environments with a heterogenous mix of servers that use different AR scripts.
`
            /* Send to a pre-defined agent */
            else if (ar_location & SPECIFIC_AGENT) {
                unsigned int i;
                char *id_list, *agent_id_str;
                ar_location++;

                for (i = 1, id_list = ar_agent_id; ; i++, id_list = NULL) {
                    agent_id_str = strtok(id_list, ",");
                    if (agent_id_str == NULL) break;

                    agent_id = OS_IsAllowedID(&keys, agent_id_str);

                    if (agent_id < 0) {
                        key_unlock();
                        merror(AR_NOAGENT_ERROR, ARGV0, agent_id_str);
                        continue;
                    }

                    send_msg((unsigned)agent_id, msg_to_send);
                }

            }
`

PR #711, opened 2015-12-10
Decoder & Rules: SSH, Clam-AV, Sysmon updated

jesuslinares:
I have been improving the following decoders and rules:
- SSH Decoder modified to extract user name when invalid/illegal users trying to log in.
- ClamAV:
  - New decoder: Extract main fields (path, virus name, hash) when a virus is detected.
  - New rule: ClamAV Stopped.
  - New rule: Virus detected multiple times.
- Sysmon Decoder:
  - New decoders for Event 1 - 8.
  - Bug fixed: "windows" must be the parent decoder of "sysmon".

We were discussing about Sysmon decoders here: https://groups.google.com/forum/#!topic/ossec-list/eAblb28kxA0

I hope it will be helpful.


ddpbsd:
The rules tests are currently failing, any chance you can fix that? It also looks like the current logs still have the archives.log headers, so removing those might be all that's necessary.

jesuslinares:
Yes, I forgot to change *contrib/ossec-testing/tests/sysmon.ini*. Anyway it is not working due to this issue https://github.com/ossec/ossec-hids/issues/712.

jesuslinares:
sysmon.ini is fixed. Travis fails due to the problem with windows decoder explained here: https://github.com/ossec/ossec-hids/issues/712.

ddpbsd:
So you don't think this is being ignored, I'm hoping to get some time this weekend to look at this and issue 712.

PR #720, opened 2015-12-23
Stop the statfs error spam

ddpbsd:


aremai:
when will this be merged? :)
or how could I fix this with my current installation manually? Will I need to recompile?

ddpbsd:
@aremai Hopefully someone will merge this, but if not you'll  have to apply the patch and recompile.

PR #727, opened 2016-01-04
Allows an optionnal ``Reply-To`` mail header

calve:
Hi all. I started writing a patch to allow adding a ``Reply-To`` header to the mails sent by Ossec.

This patch still needs testing on my side (and documentation writing), but I would like to have an early code-review in case I missed something; so here it is.

[Review on Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/727)

### DIFF
--- a/etc/ossec.conf
+++ b/etc/ossec.conf
@@ -6,6 +6,7 @@
     <email_to>daniel.cid@example.com</email_to>
     <smtp_server>smtp.example.com.</smtp_server>
     <email_from>ossecm@ossec.example.com.</email_from>
+    <!-- <email_reply_to>replyto@ossec.example.com.</email_reply_to> -->
     <picviz_output>no</picviz_output>
   </global>
 
--- a/src/config/global-config.c
+++ b/src/config/global-config.c
@@ -114,6 +114,7 @@ int Read_Global(XML_NODE node, void *configp, void *mailp)
 
     const char *xml_emailto = "email_to";
     const char *xml_emailfrom = "email_from";
+    const char *xml_emailreplyto = "email_reply_to";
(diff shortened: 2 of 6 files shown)


calve:
So this is working as expected on my side; the alerts I receive actually contains the ``Reply-to`` header. However I have no clue why the build did not pass, but it seems to fail before my changes.

Also I've committed the patch for the documentation at https://github.com/ossec/ossec-docs/pull/149

ddpbsd:
The build failures are unrelated to your submission. I've spent some time troubleshooting it, but haven't figured it out just yet (and haven't had time to contact travis-ci about it). 
Thanks for submitting the doc PR, I really appreciate it. At a glance this looks good, but I want to look it over again when I have more than a minute or two *unless someone else gets to it first!).

dcid:
Really curious here. Why you need a reply-to there? :)

Just trying to think of use cases.

calve:
My use case is the following; Ossec sends mails from ``ossecm@domain`` (which is the default) to ``alerts@domain``, which is actually an internal mailing-list in the company. 

When I reply to a mail alert and I am not fully awake, I sometime just forgot to change the recipient to ``alerts@domain`` instead of ``ossecm@domain``, and the mail is bounced back, which is not what I want. Having a ``Reply-To: alerts@domain`` header will simplify this and reduce the number of undelivered mails and missed conversations.

Last, I wanted to dig a bit in the OSSEC code ( : 

May I ask you a time window for the merge of this feature ?

jrossi:
@calve Was going to merge, but looks like this has some conflicts now.  Could you sync this branch up with master?  Will merge afterwards.

calve:
Thanks. I rebased my branch.

PR #735, opened 2016-01-26
Recalculate the permissions from octal to integer.

sleemburg:
Without this patch changes in file permissions are represented wrong (and confusing).

[Review on Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/735)

### DIFF
--- a/src/analysisd/decoders/syscheck.c
+++ b/src/analysisd/decoders/syscheck.c
@@ -430,10 +430,10 @@ static int DB_Search(const char *f_name, const char *c_sum, Eventinfo *lf)
                 }
             }
 
-            /* Get integer values */
+            /* Get octal values */
             if (c_newperm && c_oldperm) {
-                newperm = atoi(c_newperm);
-                oldperm = atoi(c_oldperm);
+                newperm = strtoul(c_newperm, 0, 8);
+                oldperm = strtoul(c_oldperm, 0, 8);
             }
 
             /* Generate size message */


dcid:
What's confusing right now? We generate/print them into the rwx/rwx/rwx format which is pretty standard.

thanks,

sleemburg:
Hi Daniel,

The problem is that they are printed wrong, because of the atoi() which
converts the octal value using base 10. Therefore a permission like 0640
get's printed like --w------T in stead of -rw-r-----

Kind regards,
Stephan

jrossi:
Sorry again about how long it has taken us to reply.  

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/analysisd/decoders/syscheck.c, line 436 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/735#-KCfxqSJ8V3OWCXt5qnM:-KCfxqSJ8V3OWCXt5qnN:-1973416749) ([raw file](https://github.com/ossec/ossec-hids/blob/4b2f0a011a7f3b28edc46afb114fe76746613b26/src/analysisd/decoders/syscheck.c#L436)):*
This is a batter way.  But don't you need to also setup errno before hand in order to check for errors? 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ossec/ossec-hids/735#-:-KCfyLOxEPfDGRCBVVL-:924840776)*

sleemburg:
Hi,
Sorry for the late reply. I know understand the confusion.

We are running a lot of agentless FIM checks. But while testing with an agent, I noticed that the results are different. I need some time to investigate, but it seems that the syscheck 'database' for agentless clients contain octal permissions, while the agent database contain decimal permissions.

So, with an agent file change, the ASCII permissions are converted correctly, but with agentless not.

Therefore, this patch is not solving the real problem. I will send a patch later which solves the real issue.

Kind regards,
Stephan

sleemburg:
I will submit another solution. The problem is that agentless checks provide octal permissions and local syschecks decimal permissions, including the filetype flags. The reporting code does not differentiate between the octal and decimal representations, so the differences for agentless modes are represented wrong and confusing.

Issue #739, opened 2016-02-02
Show ID used by the agent, when reported as invalid (not only the IP)


santiago-bassett:
Requested in the mailing list.

src/error_messages/error_messages.h:#define ENC_IP_ERROR    "%s(1408): ERROR: Invalid ID for the source ip: '%s'."

It would make troubleshooting easier. 

Best

irctrakz:
Even including the agent key/ID hash that was unable to resolve would be an awesome step forward.

santiago-bassett:
Implemented here:

https://github.com/wazuh/ossec-wazuh/commit/b277f0b159a0145d7501d446c429db19a50f922a

Vikman will do a pull request

irctrakz:
Awesome! Thank you.

PR #740, opened 2016-02-03
Modified ossec-remoted to show agent ID when reported as invalid

vikman90:
This change solves the issue https://github.com/ossec/ossec-hids/issues/739 and makes OSSEC to report the ID of the agent that tries to connect but its ID doesn't match the key.


dcid:
Can't see any drawbacks here. Looks good.

Issue #712, opened 2015-12-10
Error with syslog predecoder


jesuslinares:
Right now, windows decoder is not working:

    2013 Oct 09 17:09:04 WinEvtLog: Application: INFORMATION(1): My Script: (no user): no domain: demo1.foo.example.com: test
    
    **Phase 1: Completed pre-decoding.
           full event: '2013 Oct 09 17:09:04 WinEvtLog: Application: INFORMATION(1): My Script: (no user): no domain: demo1.foo.example.com: test'
           hostname: 'LinMV'
           program_name: 'WinEvtLog'
           log: 'Application: INFORMATION(1): My Script: (no user): no domain: demo1.foo.example.com: test'
    
    **Phase 2: Completed decoding.
           No decoder matched.

before this commit (https://github.com/ossec/ossec-hids/commit/217e6dddfd2da6887614ae59ba7ce9f48ba379f8), it was working fine:

    2013 Oct 09 17:09:04 WinEvtLog: Application: INFORMATION(1): My Script: (no user): no domain: demo1.foo.example.com: test
    
    
    **Phase 1: Completed pre-decoding.
           full event: '2013 Oct 09 17:09:04 WinEvtLog: Application: INFORMATION(1): My Script: (no user): no domain: demo1.foo.example.com: test'
           hostname: 'LinMV'
           program_name: '(null)'
           log: '2013 Oct 09 17:09:04 WinEvtLog: Application: INFORMATION(1): My Script: (no user): no domain: demo1.foo.example.com: test'
    
    **Phase 2: Completed decoding.
           decoder: 'windows'
           status: 'INFORMATION'
           id: '1'
           extra_data: 'My Script'
           dstuser: '(no user)'
           system_name: 'demo1.foo.example.com'

This is because *windows decoder* doesn't use the **program_name** tag because it was *(null)*, but now it is *WinEvtLog*.

It could be solved replacing:

    <prematch>^\d\d\d\d \w\w\w \d\d \d\d:\d\d:\d\d WinEvtLog: |^WinEvtLog: </prematch>

for

    <program_name>^WinEvtLog</program_name>

But, I think the modifications in the syslog predecoder are wrong. Why a windows event is considered as syslog event if the windows event has not Host field?:

2015 Nov 13 HH:MM:SS **Host Program:** Message
2013 Oct 09 17:09:04 **WinEvtLog:** Application: INFORMATION(1): My Script: (no user): no domain: demo1.foo.example.com: test

[...]

PR #743, opened 2016-02-17
Decoder & Rules: SSH, Clam-AV, Sysmon updated

jesuslinares:
I have been improving the following decoders and rules:
- SSH Decoder modified to extract user name when invalid/illegal users trying to log in.
- ClamAV:
  - New decoder: Extract main fields (path, virus name, hash) when a virus is detected.
  - New rule: ClamAV Stopped.
  - New rule: Virus detected multiple times.
- Sysmon Decoder:
  - New decoders for Event 1 - 8.
  - Bug fixed: "windows" must be the parent decoder of "sysmon".

We were discussing about Sysmon decoders here: https://groups.google.com/forum/#!topic/ossec-list/eAblb28kxA0

Travis fails due to the problem with windows decoder explained here: #712.
I sent the same pull request https://github.com/ossec/ossec-hids/pull/711, but I closed it due to managment issues with our repository.

I hope it will be helpful.

[Review on Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/743)


aquerubin:
On Wed, 17 Feb 2016, Jesús Linares wrote:

> I have been improving the following decoders and rules:

> - Sysmon Decoder:
>  - New decoders for Event 1 - 8.
>  - Bug fixed: "windows" must be the parent decoder of "sysmon".
>
> We were discussing about Sysmon decoders here: 
> https://groups.google.com/forum/#!topic/ossec-list/eAblb28kxA0
>
> Travis fails due to the problem with windows decoder explained here: #712.
> I sent the same pull request 
> https://github.com/ossec/ossec-hids/pull/711, but I closed it due to 
> managment issues with our repository.

Some of those decoder regex could use some simplication.  For example, in 
the Sysmon-EventID#3 decoder, you have 'DestinationIsIpv6' which is 
probably unnecessary since you already have 'DestinationIP:' anchoring 
that part of the search pattern for dstip.  Similarly, in the same regex, 
'Initiated', 'SourceHostname', 'DestinationHostname:' are probably 
unnecessary as well.  Simpler is better.


Antonio Querubin
e-mail:  tony@lavanauts.org

jesuslinares:
You are right, but sometimes the expression '\\.' with the modifier '+' doesn't work as it is expected. That is the reason I include "checkpoints" like 'DestinationIsIpv6'.

aquerubin:
On Wed, 17 Feb 2016, Jesús Linares wrote:

> You are right, but sometimes the expression '\\.' with the modifier '+' 
> doesn't work as it is expected. That is the reason I include 
> "checkpoints" like 'DestinationIsIpv6'.

It would be better to look into why it doesn't work and fix that rather 
than let regexs become more complicated than they need to be.  Do you have 
example logs that do NOT decode as you'd expect when run through the 
simpler regex?


Antonio Querubin
e-mail:  tony@lavanauts.org

jesuslinares:
Hi Antonio,

you are totally right. I haven't had time to analyse the "bug" with regex in deep. For that reason, I prefer use some "checkpoints" in the regex. Also it could be help to someone that want to extract another field. Anyway, I think it doesn't have too much impact in performance, right?.

Here you have an example, the decoder "Sysmon-EventID#3" use this regex:
`<regex offset="after_prematch">Image: (\.*)\s+User: (\.*)\s+Protocol: (\S*)\s+Initiated\.+SourceIp: (\S*)\s+SourceHostname\.+SourcePort: (\S*)\s+SourcePortName:\.+DestinationIsIpv6\.+DestinationIp: (\S*)\s+DestinationHostname:\.+\s+DestinationPort: (\S*)</regex>`

Now change it to:
`<regex offset="after_prematch">Image: (\.*)\s+User: (\.*)\s+Protocol: (\S*)\s+Initiated\.+SourceIp: (\S*)\s+SourceHostname\.+SourcePort: (\S*) \.+DestinationIp: (\S*)\s+DestinationHostname:\.+\s+DestinationPort: (\S*)</regex>`

The second regex is simpler, but it doesn't extract any field. The difference is:
```
SourcePort: (\S*)\s+SourcePortName:\.+DestinationIsIpv6\.+DestinationIp
SourcePort: (\S*) \.+DestinationIp
```

Regards.
Jesus Linares

dcid:
I have some concerns here as well.

On invalid logins, the user name is not valid. It doesn't have to be anything in the system and can actually be anything specified and controlled by a remote user (or attacker).

When we pass an invalid user to the srcuser, we are tying it up as if it was a valid system user and can lead to log injections and complications when monitoring the logs. 

Also, \.+ is evil and slow :)

jesuslinares:
I understand what you say, but in which case a field of a log could be executed?.

Also, have you test the problem I said with \.+?. I think could be a bug.

ddpbsd:
For completeness sake I like grabbing the user, even if it's not on that system. If nothing else it draws attention to that username. But it also helps reports provide more information. If reportd wasn't crashing for me, I'd try to whip up an example.

jesuslinares:
I prefer get the user too.

This PR is failing due to the issue https://github.com/ossec/ossec-hids/issues/712. Right now, windows decoders don't work on the master branch.

Issue #748, opened 2016-02-22
manage_agents on Windows exists silently


vikman90:
The program "manage_agents" doesn't work on Windows. When launching, it doesn't show any message and exists with code 0.

This worked well in v2.8.3.

PR #749, opened 2016-02-22
Changed Makefile to use Windows subsystem only wth UI manager

vikman90:
Option "-mwindows" disables console in Windows binaries. This should be used only on UI applications.

This solves the issue https://github.com/ossec/ossec-hids/issues/748.

[Review on Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/749)


jrossi:
@awiddersheim Could you take a look at this? If you have any free time?

PR #750, opened 2016-02-24
Multiple changes to logcollector config + saner defaults

dcid:
Trying to start merging my repo up.

- Better defaults to analysisd stats diff alerts.
- Fixing glob() implementation bug.
- Adding hourly/daily options to logcollector.
- Adding a CHANGELOG so we can track changes per release (a user facing CHANGELOG so they can see what matters most).

### DIFF
--- a/etc/internal_options.conf
+++ b/etc/internal_options.conf
@@ -14,11 +14,11 @@
 # Analysisd default rule timeframe.
 analysisd.default_timeframe=360
 # Analysisd stats maximum diff.
-analysisd.stats_maxdiff=25000
+analysisd.stats_maxdiff=999000
 # Analysisd stats minimum diff.
-analysisd.stats_mindiff=250
+analysisd.stats_mindiff=1250
 # Analysisd stats percentage (how much to differ from average)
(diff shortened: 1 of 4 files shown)


etc/internal_options.conf
```diff
@@ -14,11 +14,11 @@
 # Analysisd default rule timeframe.
 analysisd.default_timeframe=360
 # Analysisd stats maximum diff.
-analysisd.stats_maxdiff=25000
+analysisd.stats_maxdiff=999000
```
jrossi: You have tested this?  I am just asking to make sure but it makes sense to me gives a lot more information to users.

jrossi:
For the most part looks good to me.  Makes sense and add some great features.  

Suggestions for the future: 
* Match the style of the sources as this changes a fair amount of code just for coding style reason and is not related to change set 
* few features in a single PR

I will merge this when i get home tonight.

etc/internal_options.conf
dcid: Oh yes, been using for a while. Logs can fluctuate in size quite a bit and the original defaults were way too low.

PR #754, opened 2016-03-01
Allow the ossec users to be created on OpenBSD

ddpbsd:
I'm not sure why this worked before, or why it stopped working, but this fixes it.
This wasn't tested on Linux or any other system due to an "oops event."

[Review on Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/754)

### DIFF
--- a/src/init/adduser.sh
+++ b/src/init/adduser.sh
@@ -68,7 +68,7 @@ else
 
     for U in ${USER} ${USER_MAIL} ${USER_REM}; do
         if ! grep "^${U}" /etc/passwd > /dev/null 2>&1; then
-            ${USERADD} "${U}" -d "${DIR}" -s ${OSMYSHELL} -g "${GROUP}"
+            ${USERADD} -d "${DIR}" -s ${OSMYSHELL} -g "${GROUP}" "${U}"
         fi
     done
 fi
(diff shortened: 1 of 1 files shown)


aquerubin:
On Tue, 1 Mar 2016, Dan Parriott wrote:

> I'm not sure why this worked before, or why it stopped working, but this fixes it.
> This wasn't tested on Linux or any other system due to an "oops event."
> You can view, comment on, or merge this pull request online at:
>
>  https://github.com/ossec/ossec-hids/pull/754

I had changed the order of the adduser args because it was failing on 
certain OS but I don't recall exactly which at the moment.  I usually test 
against RHEL, Linux Mint, FreeBSD, and OpenBSD so I would think I would 
have caught this.

We may need to have a conditional in the script to check for OS and run 
the appropriate form of adduser.

Antonio Querubin
e-mail:  tony@lavanauts.org

ddpbsd:
I have updated this to only put the user at the end for OpenBSD. It should be easy enough to fiddle with if that ends up being a better default or something.

aquerubin:
On Wed, 2 Mar 2016, Dan Parriott wrote:

> I have updated this to only put the user at the end for OpenBSD. It 
> should be easy enough to fiddle with if that ends up being a better 
> default or something.

I vaguely recall I put the user at the beginning specifically because of 
fresh-install issues with OpenBSD.  But that was so long ago for either 
OpenBSD 5.6 or 5.7.  Perhaps this is no longer true for 5.8?


Antonio Querubin
e-mail:  tony@lavanauts.org

ddpbsd:
I don't remember the last time I did a fresh install, but the default failed on OpenBSD 5.9 (recent snapshot). I just retested, and this current code works. I don't currently have anything older than 5.9 to test against, but the adduser source hasn't changed much recently.

PR #755, opened 2016-03-03
Sshd rules and decoders update

ddpbsd:


ddpbsd:
I have made those changes to 5752.

PR #757, opened 2016-03-04
Refactoring the addagent files for improving the usability

netman2k:
I found that If I use the getFullnameById function directly from other files. it could not find the full name properly. cause the code tries to load the key file statically even though it uses chroot environment.

I think that it should be changed as the diff.

Please investigate this problem and merge them if you do not find any problems.

[Review on Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/757)

### DIFF
--- a/src/addagent/manage_agents.c
+++ b/src/addagent/manage_agents.c
@@ -288,9 +288,67 @@ int add_agent()
     return (0);
 }
 
-int remove_agent()
+int remove_agent_by_id(const char* id)
 {
     FILE *fp;
+    int id_exist;
+    char u_id[FILE_SIZE + 1];
+    u_id[FILE_SIZE] = '\0';
+
+    id_exist = IDExist(id);
+
+    if (!id_exist) {
+        printf(NO_ID, id);
+        return (1);
+    }
+
+    strncpy(u_id, id, FILE_SIZE);
+
+    do {
+
+        char *full_name = getFullnameById(u_id);
+
+        if (!full_name) {
+            printf(NO_ID, u_id);
+            return (1);
+        }
+
+        if (isChroot()) {
+            fp = fopen(AUTH_FILE, "r+");
+        } else {
+            fp = fopen(KEYSFILE_PATH, "r+");
+        }
+
+        if (!fp) {
+            free(full_name);
+            ErrorExit(FOPEN_ERROR, ARGV0, AUTH_FILE, errno, strerror(errno));
+        }
+#ifndef WIN32
+        chmod(AUTH_FILE, 0440);
+#endif
+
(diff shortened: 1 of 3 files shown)


netman2k:
I use the ossec-authd daemon to register new agents for automation. 
the problem is that there are too many duplicated agent record in the clients.key file.
so I started to modifying it. 
but I think some codes have to be changed to increase the usability. cause you can not call these function directly. for instance, some function needs an environment variable to remove the agent.

so I changed the addagent/manage_agents.c file as below:
1. creating a new function named remove_agent_by_id, which requires an id to remove the record.
2. modified the previous function, remove_agent, this is no longer contains remove functionality.
    instead, it uses the remove_agent_by_id function.

jrossi:
Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/addagent/manage_agents.c, line 344 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/757#-KCc7Bu48fb8YCVNAj9T:-KCc7Bu48fb8YCVNAj9U:-2039736489) ([raw file](https://github.com/ossec/ossec-hids/blob/437c3bbf24079e5c9fe17c4752a06e4f05fe39d8/src/addagent/manage_agents.c#L344)):*
Whole do look is not needed.  As you break out of it.  Should be noted that the original do loop was to allow multiple input.  This do look do not do that. 

---

*[src/addagent/manage_agents.c, line 345 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/757#-KCc7Tp9Xe6fZegIOykc:-KCc7Tp9Xe6fZegIOykd:325705443) ([raw file](https://github.com/ossec/ossec-hids/blob/f99fe94633d889a382ddd13007c699b597d8558c/src/addagent/manage_agents.c#L345)):*
Removing the Do loop means that this function and user interfaces changes.  It will no longer loop and allow many input if i am reading this correctly. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ossec/ossec-hids/757)*

atomicturtle:
Thanks @netman2k — closing without merging.

The chroot vs KEYSFILE_PATH concern is largely addressed in current addagent
(willchroot + isChroot() in IDExist/OS_RemoveAgent and friends). The
remove_agent_by_id refactor changes interactive remove UX and was blocked in
review (confirm/multi-input loop). Authd duplicate-key issues need a different
fix than this manage_agents split.

If we still need getFullnameById to use KEYSFILE_PATH outside chroot, that can
be a small follow-up matching the existing IDExist pattern.

Issue #759, opened 2016-03-06
OSSEC Invalid Signature ID


Alkaris:
I'm getting some stupid error that shouldn't be from fresh install and try to run as is after compiling it with it throwing this error;

    rules_list: Signature ID '1002' not found. Invalid 'if_sid'.

I have no idea what causes this, I haven't even touched the rules list. Agent was created, and that was fine, then after doing that I try to test run OSSEC and then I get this error for whatever reason. I've looked around for solutions online but there's nothing that helped me in figuring it out.

```
sudo /var/ossec/bin/ossec-logtest -v -d

2016/03/06 15:43:34 adding rule: rules_config.xml
2016/03/06 15:43:34 adding rule: pam_rules.xml
2016/03/06 15:43:34 adding rule: sshd_rules.xml
2016/03/06 15:43:34 adding rule: telnetd_rules.xml
2016/03/06 15:43:34 adding rule: syslog_rules.xml
2016/03/06 15:43:34 adding rule: arpwatch_rules.xml
2016/03/06 15:43:34 adding rule: symantec-av_rules.xml
2016/03/06 15:43:34 adding rule: symantec-ws_rules.xml
2016/03/06 15:43:34 adding rule: pix_rules.xml
2016/03/06 15:43:34 adding rule: named_rules.xml
2016/03/06 15:43:34 adding rule: smbd_rules.xml
2016/03/06 15:43:34 adding rule: vsftpd_rules.xml
2016/03/06 15:43:34 adding rule: pure-ftpd_rules.xml
2016/03/06 15:43:34 adding rule: proftpd_rules.xml
2016/03/06 15:43:34 adding rule: ms_ftpd_rules.xml
2016/03/06 15:43:34 adding rule: ftpd_rules.xml
2016/03/06 15:43:34 adding rule: hordeimp_rules.xml
2016/03/06 15:43:34 adding rule: roundcube_rules.xml
2016/03/06 15:43:34 adding rule: wordpress_rules.xml
2016/03/06 15:43:34 adding rule: cimserver_rules.xml
2016/03/06 15:43:34 adding rule: vpopmail_rules.xml
2016/03/06 15:43:34 adding rule: vmpop3d_rules.xml
2016/03/06 15:43:34 adding rule: courier_rules.xml
2016/03/06 15:43:34 adding rule: web_rules.xml
2016/03/06 15:43:34 adding rule: web_appsec_rules.xml
2016/03/06 15:43:34 adding rule: apache_rules.xml
[...]

PR #762, opened 2016-03-07
Tweak the sshd rules and tests

ddpbsd:
Change one if_sid from 1002 to 5700, and make some corrections to the tests.


ddpbsd:
This should fix issue #759

PR #765, opened 2016-03-08
Do not show diff

calve:
Sensitive data may leak through the diff attached to alerts when some file changes. This pull request add a ``nodiff`` option, which allows to explicitely set files for which we never want to output a diff.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/765)

### DIFF
--- a/src/config/syscheck-config.h
+++ b/src/config/syscheck-config.h
@@ -59,6 +59,9 @@ typedef struct _config {
     char **ignore;                  /* list of files/dirs to ignore */
     OSMatch **ignore_regex;         /* regex of files/dirs to ignore */
 
+    char **nodiff;                  /* list of files/dirs to never output diff */
+    OSMatch **nodiff_regex;         /* regex of files/dirs to never output diff */
+
     char **dir;                     /* array of directories to be scanned */
     OSMatch **filerestrict;
 
--- a/src/syscheckd/seechanges.c
+++ b/src/syscheckd/seechanges.c
@@ -41,6 +41,11 @@ int is_text(magic_t cookie, const void *buf, size_t len)
 }
 #endif
 
+/* Return 1 if the file has the ``nodiff`` option. 0 otherwise */
+int is_nodiff(const char *filename){
+    return 0;
+}
+
 /* Generate diffs alerts */
 static char *gen_diff_alert(const char *filename, time_t alert_diff_time)
 {
@@ -347,15 +352,25 @@ char *seechanges_addfile(const char *filename)
         goto cleanup;
     }
 
-    /* Run diff */
-    snprintf(
-        diff_cmd,
-        2048,
-        "diff \"%s\" \"%s\" > \"%s\" 2> /dev/null",
-        new_tmp,
-        old_tmp,
-        diff_tmp
-    );
+    if (is_nodiff((filename))) {
+        /* Dont leak sensible data with a diff hanging around */
+        snprintf(
+            diff_cmd,
+            2048,
+            "echo diff truncated > \"%s\" 2> /dev/null",
+            diff_tmp
+        );
+    } else {
+        /* OK, run diff */
+        snprintf(
+            diff_cmd,
+            2048,
+            "diff \"%s\" \"%s\" > \"%s\" 2> /dev/null",
+            new_tmp,
+            old_tmp,
+            diff_tmp
+        );
+    };
 
     if (system(diff_cmd) != 256) {
         merror("%s: ERROR: Unable to run diff for %s", ARGV0, filename);
--- a/src/syscheckd/seechanges.c
+++ b/src/syscheckd/seechanges.c
@@ -43,6 +43,15 @@ int is_text(magic_t cookie, const void *buf, size_t len)
 
 /* Return 1 if the file has the ``nodiff`` option. 0 otherwise */
 int is_nodiff(const char *filename){
+    int i = 0;
+    if (syscheck.nodiff){
+        for (i = 0; syscheck.nodiff[i] != NULL; i++){
+            if (strncasecmp(syscheck.nodiff[i], filename,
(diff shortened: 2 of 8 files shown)


jrossi:
This is great and a very useful feature.   While we review could add this to the docs: https://github.com/ossec/ossec-docs by creating a pull request?

src/syscheckd/seechanges.c
```diff
@@ -347,23 +361,41 @@ char *seechanges_addfile(const char *filename)
...
-        new_tmp,
-        old_tmp,
-        diff_tmp
-    );
+    if (is_nodiff((filename))) {
+        /* Dont leak sensible data with a diff hanging around */
+        snprintf(
+            diff_cmd,
+            2048,
+            "printf '<Diff truncated>' > \"%s\"",
```
jrossi: Think this could be more descriptive with something like Diff truncated due to nodiff

jrossi: Also is printf always present on all supported platforms.  I really hate shelling out for this use case it's pointless outside of integrating into the reset of the function.

src/config/syscheck-config.h
```diff
@@ -59,6 +59,9 @@ typedef struct _config {
     char **ignore;                  /* list of files/dirs to ignore */
     OSMatch **ignore_regex;         /* regex of files/dirs to ignore */
 
+    char **nodiff;                  /* list of files/dirs to never output diff */
+    OSMatch **nodiff_regex;         /* regex of files/dirs to never output diff */
```
jrossi: Nodiff_regex does not look used.  Maybe I am missing something.

jrossi:
Also it's not clear that function is_nodiff checks the osman chest if that was used.

src/syscheckd/seechanges.c
calve: Are you suggesting to directly fwrite() to the file ?

jrossi: Nope, sorry for not being clear.  Given the code this is sane solution.

jrossi:
One thing that I don't see is how OSmatch is ever used.  The is_nodiff function only processes the list of files but does not process the OSmatches that are created via the Conf/xml

jrossi:
Reviewed 8 of 8 files at r1.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[src/config/syscheck-config.c, line 714 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/765#-KCc8bXUnd7oPYR7UFjJ:-KCc8bXV0ueE7OOLWemU:1225474886) ([raw file](https://github.com/ossec/ossec-hids/blob/e52ca36f66b96fdddb59da68da23d52af609a518/src/config/syscheck-config.c#L714)):*
This can fail: 

```
If the function fails, the return value is zero. To get extended error information, call GetLastError.
```
> https://msdn.microsoft.com/en-us/library/windows/desktop/ms724265(v=vs.85).aspx

What happens if it fails? 

---

*[src/syscheckd/seechanges.c, line 52 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/765#-KCc9YOAbocbYuQSGllh:-KCc9YOAbocbYuQSGlli:-1117939194) ([raw file](https://github.com/ossec/ossec-hids/blob/e52ca36f66b96fdddb59da68da23d52af609a518/src/syscheckd/seechanges.c#L52)):*
Yeah more closer review shows no regex/osmatch testing is done.  Could we get this added or remove the feature from the config section.   

---

*[src/syscheckd/syscheck.c, line 144 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/765#-KCc9rt2qLajI4aGPhdv:-KCc9rt2qLajI4aGPhdw:-893206127) ([raw file](https://github.com/ossec/ossec-hids/blob/e52ca36f66b96fdddb59da68da23d52af609a518/src/syscheckd/syscheck.c#L144)):*
We try to use {} even if not required as it leads to problems later (sometimes not always). 

---

*[src/syscheckd/syscheck.c, line 326 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/765#-KCcA-5tqAjGPVyok9g_:-KCcA-5tqAjGPVyok9ga:-893206127) ([raw file](https://github.com/ossec/ossec-hids/blob/e52ca36f66b96fdddb59da68da23d52af609a518/src/syscheckd/syscheck.c#L326)):*
We try to use {} even if not required as it leads to problems later (sometimes not always). 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ossec/ossec-hids/765)*

calve:
Thanks for the review. I updated the code according to your suggestions. I added the regex match feature.

As for ``ExpandEnvironmentStrings()`` failures, I have absolutely no idea what to do. The rest of the code base does not seems to check the return code either. I would suggest to drop that Windows-only block, as diff are not runned on these systems (per the documentation). On the other side, maybe someone will port the feature some days.

jrossi:
For ExpandEnvironmentStrings good point I did not check on that.  Thank you.  

---

Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ossec/ossec-hids/765#-:-KCforK1MV1urAKyPD1a:-1761596150)*

jrossi:
@calve Could you add docs github.com/ossec/ossec-docs ?  Also add yourself as contributor?  

Thank you again great stuff and something I think others will use.

calve:
Hi guys,

isn't it worth assigning a CVE under Information Leakage (https://cwe.mitre.org/data/definitions/200.html) to spread the fact that sensible informations might have leaked over email ? cc @ddpbsd

PR #770, opened 2016-03-10
Fixed some PostgreSQL issues

teaearlgraycold:
As for what this fixes:

* The user column needs to be escaped else Postgres will think it's the `USER` keyword. This was acknowledged in the schema file but not in the alert insertion statement in the C code.
* The schema file has some conflicting index names so I prefixed each with the name of the table it's for
* Many alerts do not match the schema defined, as they don't provide a `user` or `src_ip`/`dst_ip`, so I removed the `NOT NULL` constraints from those columns on the alerts table.

---

There should also probably be a prompt in the `install.sh` script to ask users if they want Postgres or MySQL support in the application.

I didn't want to mess around with that behemoth of a script though.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/770)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -101,7 +101,9 @@ Install()
 
     # Binary install will use the previous generated code.
     if [ "X${USER_BINARYINSTALL}" = "X" ]; then
-        ${MAKEBIN} PREFIX=${INSTALLDIR} TARGET=${INSTYPE} build 
+        # Add DATABASE=pgsql or DATABASE=mysql to add support for database
+        # alert entry
+        ${MAKEBIN} PREFIX=${INSTALLDIR} TARGET=${INSTYPE} build
         if [ $? != 0 ]; then
             cd ../
             catError "0x5-build"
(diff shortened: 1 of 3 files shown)


teaearlgraycold:
I should also add that these changes were made with PostgreSQL 9.4 in mind.

jrossi:
:lgtm:

---

Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ossec/ossec-hids/770#-:-KCg0bqpVpMTfMsXSmaf:1417088721)*

atomicturtle:
We might need to revert this one, it is breaking mysql compatibility on :
             "alert(server_id,rule_id,level,timestamp,location_id,src_ip,src_port,dst_ip,dst_port,alertid,\"user\",full_log) "

\"user\" is not valid syntax

teaearlgraycold:
Should add a check for MySQL/PGSQL and use different query strings

teaearlgraycold:
Made a new PR:

https://github.com/ossec/ossec-hids/pull/804

PR #771, opened 2016-03-10
Setting up GeoIP by default

dcid:
Bit of a big pull request, but set ups GeoIP by default (hard to do a small PR since it touches multiple parts).

1- Auto installs / downloads the maxmind db on install.sh

2- Always does a geo lookup whenever a srcip /dstip is found.

3- Always includes the Geo data within the alert, like:

```Rule: 5715 (level 3) -> 'SSHD authentication success.'
Src IP: XXX / USA / New Jersey
User: root
Mar 10 17:36:36 test-dcid-ossec sshd[15538]: Accepted publickey for root from XXX port 51393````

4- Add <srcgeoip> and <dstgeoip> to the rules, so you can create granular rules based on the country or city.

5- Adds libgeoip to the external folder, so always compiled.

Have been running this code in production for many months and working well.

<!-- Reviewable:start -->
---
This change is [<img src="https://reviewable.io/review_button.svg" height="35" align="absmiddle" alt="Review on Reviewable"/>](https://reviewable.io/reviews/ossec/ossec-hids/771)
<!-- Reviewable:end -->


ddpbsd:
Can you change the .travis.yaml file to set GEOIP to "yes" for each item? I think that should cause the checks to pass.

jrossi:
Reviewed 2 of 24 files at r1.
Review status: 2 of 24 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[install.sh, line 756 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/771#-KCaCboxqvkUedNi6AI7:-KCaCboxqvkUedNi6AI8:-1861862083) ([raw file](https://github.com/ossec/ossec-hids/blob/b04febf8b2639a3b8907c614251515427c29301b/install.sh#L756)):*
I love the single make things work for users.  But I do get worried about downloading directly and not testing for failures.  In highly proxied network this would fail horribly.  

---

*[install.sh, line 1221 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/771#-KCay_uaC_f7XyennSRb:-KCay_uaC_f7XyennSRc:59683500) ([raw file](https://github.com/ossec/ossec-hids/blob/b04febf8b2639a3b8907c614251515427c29301b/install.sh#L1221)):*
I don't mind the code and auto installing the included lib, but adding the MaxMind DB by default might not be the best.  Or at least have be on by default and then options to disable.  

---

*[src/analysisd/decoders/decoder.c, line 263 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/771#-KCaBrn2r1iGeRyVsS8z:-KCaBrn2r1iGeRyVsS9-:768590303) ([raw file](https://github.com/ossec/ossec-hids/blob/b04febf8b2639a3b8907c614251515427c29301b/src/analysisd/decoders/decoder.c#L263)):*
Bare IF have hurt OSSEC in the past.  We are attempting to remove them where possible as makes reading and logics less easy to understand. 

---

*[src/analysisd/decoders/decoder.c, line 276 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/771#-KCaCAcmAlA-6uQjzoFI:-KCaCAcmAlA-6uQjzoFJ:630493574) ([raw file](https://github.com/ossec/ossec-hids/blob/b04febf8b2639a3b8907c614251515427c29301b/src/analysisd/decoders/decoder.c#L276)):*
Bare IF without braces. 

---

*[src/analysisd/decoders/geoip.c, line 41 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/771#-KCaz6esNdA7147Kn068:-KCaz6esNdA7147Kn069:-1417921176) ([raw file](https://github.com/ossec/ossec-hids/blob/b04febf8b2639a3b8907c614251515427c29301b/src/analysisd/decoders/geoip.c#L41)):*
Does this work with IPv6 cleanly?   I read some of the code but it was not completely clear. 

---

*[src/external/libgeoip/Makefile, line 8 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/771#-KCazpmd9U7koBxLw0UO:-KCazpmezo-w6SZbxLpx:-375607480) ([raw file](https://github.com/ossec/ossec-hids/blob/b04febf8b2639a3b8907c614251515427c29301b/src/external/libgeoip/Makefile#L8)):*
This is not used in the current build system.  Config.Make is gone. 

---

*[src/external/libgeoip/Makefile, line 11 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/771#-KCazyk308fSUMlSFU3g:-KCazyk308fSUMlSFU3h:1448649118) ([raw file](https://github.com/ossec/ossec-hids/blob/b04febf8b2639a3b8907c614251515427c29301b/src/external/libgeoip/Makefile#L11)):*
package version looks like something from other code base. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ossec/ossec-hids/771)*

jrossi:
I was just trying out Reviewable for this large commit.

jrossi:
Reviewed 22 of 24 files at r1.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[src/analysisd/analysisd.c, line 1338 \[r1\]](https://reviewable.io:443/reviews/ossec/ossec-hids/771#-KCb0650WIAmzQSyd73x:-KCb0650WIAmzQSyd73y:-442143620) ([raw file](https://github.com/ossec/ossec-hids/blob/b04febf8b2639a3b8907c614251515427c29301b/src/analysisd/analysisd.c#L1338)):*
Would using cdb looks for data of this nature be useful?  Not really something that is required in this PR.  But something to think about. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ossec/ossec-hids/771)*

aquerubin:
On Thu, 10 Mar 2016, Daniel B. Cid wrote:

> Bit of a big pull request, but set ups GeoIP by default (hard to do a 
> small PR since it touches multiple parts).
>
> 1- Auto installs / downloads the maxmind db on install.sh
>
> 2- Always does a geo lookup whenever a srcip /dstip is found.
>
> 3- Always includes the Geo data within the alert, like:
>
> ```Rule: 5715 (level 3) -> 'SSHD authentication success.'
> Src IP: XXX / USA / New Jersey
> User: root
> Mar 10 17:36:36 test-dcid-ossec sshd[15538]: Accepted publickey for root from XXX port 51393````
>
> 4- Add <srcgeoip> and <dstgeoip> to the rules, so you can create 
> granular rules based on the country or city.
>
> 5- Adds libgeoip to the external folder, so always compiled.
>
> Have been running this code in production for many months and working well.
>
>
> You can view, comment on, or merge this pull request online at:
>
>  https://github.com/ossec/ossec-hids/pull/771
>
> -- Commit Summary --
>
>  * -Adding libgeoip by default.
>  * Moving maxmind db automatically to /var/ossec.
>  * Second batch to merge code to enable geoip by default.
>  * Missing geoip.c file.
>  * Fixing Makefile for GeoIP
>  * Printing geoip into the logs by default.

Do we really want to default this feature to 'on'?  While it's nice to 
have the info as to where the attacks are coming from, or in the case of a 
distributed enterprise, which of your data centers are under sustained 
attack, that information doesn't necessarilly help you mitigate the 
attack.  This is really more an analysis presentation function.

I think it would be better to use calls to a packaged shared-library than 
incorporate the geoip source into the ossec source.  The geoip lib is 
already available as individual packages for many systems.  The main 
difficulty is ensuring the library can be called in a chroot environment 
but that may be just a simple matter of creating /var/ossec/lib and 
copying the necessary lib and data files there.  Also, the API source code 
as incorporated in the PR is a disaster when it comes to handling both 
IPv4 and IPv6 addresses.  Rather than generalizing the IPv4-specific 
functions and pushing the version-specific code down to the lowest level, 
it seems they've been replicated and modified for IPv6.  Yes it may work 
but OMG that's ugly :)


Antonio Querubin
e-mail:  tony@lavanauts.org

dcid:
@aquerubin I think it should to be on by default. I mean, it can be easily disabled if you remove the entry for the geoipdb on ossec.conf.

Operationally speaking, correlating geoip information is a lot more useful than just IP addresses and opens up the doors for many interesting rules by default (eg: alert for 2 successful SSH logins from different countries from same user within few hours).

If we believe OSSEC should be useful by default, the majority of users will benefit of it.

aquerubin:
On Fri, 11 Mar 2016, Daniel B. Cid wrote:

> @aquerubin I think it should to be on by default. I mean, it can be 
> easily disabled if you remove the entry for the geoipdb on ossec.conf.

And it can just as easily be enabled then.

> Operationally speaking, correlating geoip information is a lot more 
> useful than just IP addresses and opens up the doors for many 
> interesting rules by default (eg: alert for 2 successful SSH logins from 
> different countries from same user within few hours).

No argument with that.  However I believe defaults should be set based on 
on the most common use cases.  At some point you run into the law of 
diminishing returns.

> If we believe OSSEC should be useful by default, the majority of users 
> will benefit of it.

I concur but as I questioned in an earlier reply, is OSSEC the right place 
to put this functionality?


Antonio Querubin
e-mail:  tony@lavanauts.org

dcid:
@aquerubin Yes, where else would you put it? How would you create rules and correlations that also look at the geoip if OSSEC does not support it?

I also think that most people do not use it because we do not enable it by default. The moment it becomes and we ship with useful default rules that make use of it, you will see the difference :)

Plus, you anyone can disable the feature if they do not want it.

thanks,

lostinthetubez:
I think this would make sense to put in the install.sh script as an option, just like how active response is enabled or disabled during an install. Users should at the very least be aware when an application is going to make network connections to outside entities. I think it makes a lot of sense to have GeoIP lookups in OSSEC. Yes, traditionally they are also conducted at the SIEM level, but when you include them in OSSEC you can use it in tandem with active response, which seems like a pretty valuable feature.

aquerubin:
On Tue, 15 Mar 2016, lostinthetubez wrote:

> I think this would make sense to put in the install.sh script as an 
> option, just like how active response is enabled or disabled during an 
> install. Users should at the very least be aware when an application is 
> going to make network connections to outside entities. I think it makes 
> a lot of sense to have GeoIP lookups in OSSEC. Yes, traditionally they 
> are also conducted at the SIEM level, but when you include them in OSSEC 
> you can use it in tandem with active response, which seems like a pretty 
> valuable feature.

I have no argument with that.  However, until the feature is mature enough 
where you can write a set of rules that takes advantage of it to trigger 
an actual alert or active response, I'd say flag it as experimental in the 
install script and default it to off.  The install script or documentation 
should also alert the installer that the database will require periodic 
updates, etc.

There is still the question of whether to include all the additional 
source code or just the calls to a shared lib.  The latter would mean 
additional development work for the PR but would maintain a cleaner 
separation between the ossec code and the geoip code.

Antonio Querubin
e-mail:  tony@lavanauts.org

ddpbsd:
I think being able to get rid of some of the ifdef mess in the code would be worth requiring geoip. But I also don't agree with bundling the libgeoip code, let the package manager handle that. Also, I don't think the installer should automatically download the latest database. An option or even a warning would be ok.

mstarks01:
GeoIP is a really, really useful feature in SIEMs and I have been using it in OSSEC since Xavier Mertens originally submitted the GeoIP code. Having it in rules is awesome. Here's one real scenario where I would have loved to have it in a rule and AR: I once downloaded an Android email app that then resulted in successful IMAP logins from Russia. I could have preemptively blocked any login outside of my metropolitan area with something like this. Another one that would be really useful is to be able to use hashes in rules, from a CDB lookup. IMHO, IDSes these days have to have contextual awareness to be really useful.

ddpbsd:
@dcid Why did you close this? I don't think we should give up on this functionality.

PR #774, opened 2016-03-15
Add the opensmtpd files to the install

ddpbsd:
This allows tests to pass. It doesn't look like the rules files are being updated/installed during an upgrade though,


ddpbsd:
This should fix the travis-ci checks that are currently failing for everyone. If we'd rather, I can remove the opensmtpd checks instead.

ChristianBeer:
This should be merged ASAP to get rid of the failed status image and make sure other pull requests are checked properly.

Issue #780, opened 2016-03-23
Compile warning (and potential segfault) after merge from calve/do_not_show_diff


vikman90:
When I compile the master branch I get the following warning:

```
config/syscheck-config.c: In function ‘Read_Syscheck’:
config/syscheck-config.c:751:21: warning: format ‘%s’ expects argument of type ‘char *’, but argument 2 has type ‘char **’ [-Wformat=]
                     verbose("Set nodiff regex: %s", syscheck->nodiff_regex[nodiff_size]->patterns);
```

Indeed, ```syscheck->nodiff_regex[]->patterns``` is a pointer to a matrix and might raise a segmentation fault if was read as a string.

This issue is caused by the commit https://github.com/ossec/ossec-hids/commit/d8a57c4024519fb6f51be3e6f978519ef6d759fd (pull request https://github.com/ossec/ossec-hids/pull/765).

vikman90:
Issue solved with merge https://github.com/ossec/ossec-hids/pull/781. Thanks!

PR #781, opened 2016-03-23
Replaced verbose for debug1 at syscheck-config.c

vikman90:
Also removed printing of a pattern matrix that might raise a segfault. This shouldn't matter because the pattern is printed just before.

This commit solves the issue https://github.com/ossec/ossec-hids/issues/780.


calve:
:+1:

Issue #785, opened 2016-03-28
SSHD rules error


jesuslinares:
In this commit https://github.com/ossec/ossec-hids/commit/43a9a279ac103001aedb49e7ed1d19b5ca646740 new ssh rules were added.

The test sshd.ini didn't work, so it was fixed using this https://github.com/ossec/ossec-hids/commit/4b4569a5f8834852a743c0938d3d18df85011cef

I think, the first sshd.ini was fine, but it is necesary to change the alert level of rule 5752 from 0 to 1.

Change:
`<rule id="5752" level="0">`

to:
`<rule id="5752" level="1">`

This way, the first sshd.ini works:

```
[ssh no matching key exchange]
log 1 pass = Sep 16 05:46:56 junction sshd[1961]: fatal: Unable to negotiate with 108.229.36.174: no matching key exchange method found. Their offer: diffie-hellman-group1-sha1,diffie-hellman-group-exchange-sha1 [preauth]

rule = 5750
alert = 2
decoder = sshd
```

ddpbsd:
We can have it match 5750 by dropping that to level 0 as well. Do we want to see that alerts?

jesuslinares:
This log:

> Sep 16 05:46:56 junction sshd[1961]: fatal: Unable to negotiate with 108.229.36.174: no matching key exchange method found. Their offer: diffie-hellman-group1-sha1,diffie-hellman-group-exchange-sha1 [preauth]

Rule 5752 with level **0**:

```
**Phase 1: Completed pre-decoding.
       full event: 'Sep 16 05:46:56 junction sshd[1961]: fatal: Unable to negotiate with 108.229.36.174: no matching key exchange method found. Their offer: diffie-hellman-group1-sha1,diffie-hellman-group-exchange-sha1 [preauth]'
       hostname: 'junction'
       program_name: 'sshd'
       log: 'fatal: Unable to negotiate with 108.229.36.174: no matching key exchange method found. Their offer: diffie-hellman-group1-sha1,diffie-hellman-group-exchange-sha1 [preauth]'

**Phase 2: Completed decoding.
       decoder: 'sshd'
       srcip: '108.229.36.174'

**Phase 3: Completed filtering (rules).
       Rule id: '5752'
       Level: '0'
       Description: 'sshd could not negotiate with client.'

```

Rule 5752 with level **1**:

```
**Phase 1: Completed pre-decoding.
[...]

PR #786, opened 2016-03-28
Rework the sshd rules to make a bit more sense.

ddpbsd:
5752 -> 5750 - This seems like a better catch-all
5750 -> 5752 - Making this rely on the "new" 5750 is cleaner (more similar to 5753).
5753 now if_sids 5750

Prompted by @jesuslinares in Issue #785


colinmollenhour:
On one of my Ubuntu 12.04 systems (OpenSSH_5.9p1...) an error that isn't caught by these new SSH rules goes like this:

> Rule: 1002 fired (level 2) -> "Unknown problem somewhere in the system."
> Portion of the log(s): Apr 18 21:27:08 web2 sshd[23484]: fatal: Unable to negotiate a key exchange method [preauth]

ddpbsd:
@colinmollenhour Thanks! I think PR #805 should fix it

PR #789, opened 2016-03-30
Use tests instead of ls to check for file existence

mmb:


dcid:
That's not portable outside of bash, unfortunately.

mmb:
If the target is Bourne shell those should work (see https://en.wikibooks.org/wiki/Bourne_Shell_Scripting/Appendix_C:_Quick_Reference#File_tests)

That script already uses the -f test here: https://github.com/ossec/ossec-hids/blob/master/install.sh#L1227

dcid:
I would actually change that other one to get them all in sync using ls :)

But the main question is: what issue are you trying to solve? The current format is very portable and works with bash and old sh. That's why we chose this way to start with. 

*note that it might be ok to change, but I don't have access to solaris/hpux/aix, like I used before, to test. These along with some *BSD's were the problem and forced us to use ls for portability.

ddpbsd:
It looks like your tree is a bit behind MASTER. Could you update (hoping to get the tests to pass)?

mmb:
I rebased against master and the test went green.

ddpbsd:
Anyone with one of these niche systems want to give this a spin?

ddpbsd:
I have a rant about AIX/HPUX institutions, but it's not appropriate here. Since there are no reports of this breaking anything (and I doubt there will be until after a release), I'm merging this. Thanks for the PR @mmb, and I expect to hear an "I told you so" @dcid.

Issue #675, opened 2015-10-02
integrate /boot into the syscheck


aremai:
please integrate /boot into the directories to check for the syscheck module.
i think it's important to also scan /boot

thanks,
t.

aremai:
any reason why this is still open?
if it's any easier, I'm happy to send a PR with the updated ossec.conf ...

aremai:
thank you @ddpbsd :)

ddpbsd:
No problem. I think further discussion can/should happen in the PR, so I'll be closing this issue.

PR #791, opened 2016-04-01
Syscheck boot

ddpbsd:
Add /boot to default syscheck directories. I don't think this should have any ill effects on non-linux systems.
This solves issue #675


aremai:
interesting! didn't know this needed changing in so many files...

looks like, i learn something new every day ;)

ddpbsd:
I don't know if all of those changes are necessary, but the situation is pretty bad. If someone has a particular fondness for shell scripting, please save us from ourselves!

Issue #752, opened 2016-02-26
Missing Sanity Checks/Null Pointer Dereference (CWE-476)/Memory Leaks in rootcheck


dogbert2:
Hello All,

	In reviewing calls to malloc() in the latest version of rootcheck, I found
numerous instances where return values for malloc are not checked for a value
of NULL, indicating failure.  Additionally, calls to memset() and strncpy()
are made immediately afterwards, but if the return value from malloc() is
NULL, these calls will result in a segmentation violation/fault.

Additionally, previously allocated memory is also not released in the
event that malloc() fails.

The three patch files below should address/correct these issues:

```
--- GeoIP.c.orig	2016-02-26 12:39:26.334682615 -0800
+++ GeoIP.c	2016-02-26 12:43:29.717615334 -0800
@@ -775,6 +775,8 @@
 {
     if (NULL == GeoIPDBFileName) {
         GeoIPDBFileName = malloc(sizeof(char *) * NUM_DB_TYPES);
+	if (GeoIPDBFileName == NULL)
+	    return;
         memset(GeoIPDBFileName, 0, sizeof(char *) * NUM_DB_TYPES);
 
         GeoIPDBFileName[GEOIP_COUNTRY_EDITION] = _GeoIP_full_path_to(
@@ -2387,6 +2389,9 @@
         } else {
             len = sizeof(char) * (strlen(buf) + 1);
             org_buf = malloc(len);
+	    if (org_buf == NULL) {
+		return NULL;
+	    }
             strncpy(org_buf, buf, len);
         }
     } else {
@@ -2396,6 +2401,9 @@
         } else {
             len = sizeof(char) * (strlen(buf_pointer) + 1);
             org_buf = malloc(len);
+	    if (org_buf == NULL) {
+		return NULL;
+	    }
             strncpy(org_buf, buf_pointer, len);
         }
     }
@@ -2447,6 +2455,9 @@
         } else {
             len = sizeof(char) * (strlen(buf) + 1);
             org_buf = malloc(len);
+	    if (org_buf == NULL) {
+		return NULL;
+	    }
             strncpy(org_buf, buf, len);
         }
     } else {
@@ -2456,6 +2467,9 @@
         } else {
             len = sizeof(char) * (strlen(buf_pointer) + 1);
             org_buf = malloc(len);
+	    if (org_buf == NULL) {
+		return NULL;
+	    }
             strncpy(org_buf, buf_pointer, len);
         }
     }
[...]

PR #792, opened 2016-04-01
Check for a NULL from malloc.

ddpbsd:
I'm not sure if returning 0 is the best option here, just copied it from issue #752


ddpbsd:
I think I got the return right in c63b476, but I'm not totally positive.

dcid:
I always like to add some logging on memory failures, but tricky to do on some cases. Seems ok here.

It might be better long term to have a saved buffer we can release on memory failure error, just so we can log things properly before exiting. Unrelated to this :)

ddpbsd:
I can add some merrors to the failures, that shouldn't be too hard.

ddpbsd:
I added some merrors, although I didn't do anything to help specify which one was erroring out. Not sure if it matters at that point?

PR #797, opened 2016-04-08
removed old rootcheck options (daemon and notify)

aremai:
removed old rootcheck options daemon and notify in the rootcheck-config.c and rootcheck.conf -- please check if everything still works as expected


ddpbsd:
I just installed this, so I'll keep an eye on it for a bit to see how things go.

aremai:
awesome, thanks mucho Dan!

PR #800, opened 2016-04-13
Add decoders, rules, and tests for OpenBSD's doas

ddpbsd:


ddpbsd:
I have no idea why this is failing. "It works for me."

PR #801, opened 2016-04-13
Remove legacy systemd files, consolidate this on ossec-control in a single .service file

atomicturtle:
This maintains parity with systemctl, service, and calling ossec-control
directly. The previous implementation would not support direct integration with ossec-control, and would launch processes out of order.


ddpbsd:
Anyone with systemd test this?

bbayles:
On the latest Ubuntu 16.04 build this does fine with `systemctl stop`, `systemctl start`, and `systemctl reload`.

The install script does not put the unit file into `/etc/systemd/system` or `/lib/systemd/system`, so the service must be installed manually.

atomicturtle:
We didnt have the previous implementation in the install.sh either so its probably better to resolve that in a separate PR than piling it on to this one

phemmer:
👎 
This is opposite of how systemd is supposed to be used. Oneshot services was meant to handle commands that run and then exit. This causes problems as it breaks thing such as the ability for systemd to restart the service if it fails.

atomicturtle:
By all means if you have a solution for that, please send us a pull request

ddpbsd:
So the way I see it (from the outside as a non-systemd user):

- ossec-control and systemctl are currently mutually exclusive

- systemd is starting things out of order

- The install script is currently clueless as to the entirety of systemd

Possible fix for the first item is updating the documentation to refer to systemctl as well as ossec-control, or provide the facilities in ossec-control to realize the system is controlled by systemd and utilize that.

Possible fix for the second is to adjust the service files to properly account for run order. [The documentation](https://www.freedesktop.org/software/systemd/man/systemd.unit.html) makes me think this wouldn't be too difficult, but that could be wishful thinking.

And lastly, adjusting `install.sh` is simple-ish. Probably more simple than other current issues with it.

Is this off base? Am I missing something important?

bbayles:
I'd like to be able to control OSSEC services more naturally with systemd init, and agree with @phemmer that the way OSSEC currently expects services to be controlled leaves something to be desired. However, I think the changes in this PR are an improvement to the status quo (and don't affect the normal install anyway) and should be merged.

phemmer:
> atomicturtle wrote:
> By all means if you have a solution for that, please send us a pull request

I've been using the current systemd files for several months without issue.

> ddpbsd wrote:
> Possible fix for the first item is updating the documentation to refer to systemctl as well as ossec-control, or provide the facilities in ossec-control to realize the system is controlled by systemd and utilize that.

Those were my thoughts as well.
The ossec-control thing is rather uncommon. Most services are managed through the init system. systemd just forces the issue as it can't control services that weren't started through it like other init systems can.

> Possible fix for the second is to adjust the service files to properly account for run order. The documentation makes me think this wouldn't be too difficult, but that could be wishful thinking.

Setting the order of systemd service is very easy. I might be willing to put up a PR for that if I knew what the order was. But as I mentioned earlier, the current services have been working fine for me. Though if they are really order dependent, it could be a simple race condition, and they happen to start in the right order on my systems.

ddpbsd:
Getting the order right isn't difficult, just have to juggle a bit of info. I have almost everything done, and I'll be submitting. PR soon for that. But it's not a major priority till after 2.9

PR #802, opened 2016-04-14
Allow for longer entries in the system.information column

infracaninophile:
Switch from VARCHAR(128) to TEXT to allow for longer entries in the
information column.  On server registration, this column is typically
loaded with the output of ```uname -a```.  That's frequently longer than
128 characters.  For instance on the FreeBSD servers I manage, it's
typically in excess of 180 characters:

```
ox-dell39:~:% uname -a | wc
       1      15     185
```

Since the length of the data being written to that column is checked,
this results in a failed SQL query at which the ossec-dbd process quits.

Tested with postgresql; equivalent change applied to mysql.schema


ddpbsd:
How does this change affect current installations/upgrades?

infracaninophile:
On 04/14/16 13:00, Dan Parriott wrote:
> How does this change affect current installations/upgrades?

This should only ever affect new installs.  Anyone already using the
original schema successfully has no need to update it.

PR #803, opened 2016-04-14
More sshd decoders

ddpbsd:
Found some logs, found similar logs on the internet, tried them out.
These offer more decoded bits for some logs. Some of the logs return 5711, but there is a username available. I haven't done more investigation into these yet to find an example log without a username.


santiago-bassett:
thank you Dan

PR #804, opened 2016-04-14
Use different alert query for MySQL and PostgreSQL

teaearlgraycold:
In response to @atomicturtle on this PR: https://github.com/ossec/ossec-hids/pull/770

### DIFF
--- a/src/os_dbd/alert.c
+++ b/src/os_dbd/alert.c
@@ -9,6 +9,7 @@
 
 #include "dbd.h"
 #include "config/config.h"
+#include "config/dbd-config.h"
 #include "rules_op.h"
 
 /* Prototypes */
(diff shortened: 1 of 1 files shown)


ddpbsd:
Anyone use ossec-dbd that wants to test this?

atomicturtle:
Yeah I've got the test environment for this one. I'll get to it shortly

atomicturtle:
I'm not having any success getting this to work with mysql, it keeps defaulting to the postgres case for some reason

teaearlgraycold:
I honestly wouldn't be surprised if that dbtype flag isn't being set correctly. The dbd module doesn't feel very fleshed out.

atomicturtle:
Right now this is a release blocking issue & we may need to revert the postgres commit

atomicturtle:
Im going to close this one out, I re-implemented this PR in #824 with the break; fix

Thanks again for the PR!

PR #819, opened 2016-05-02
2 9 b Modify df -h command

Pantera1989m:
The -h flag causes the output of df to be multiline if the path is very long and the output is not parsed correctly by OSSEC. I propose the command to be changed to "df -Pkh" which will cause the output to be in a single line. The rule 531 still works by simply modifying "<match>ossec: output: 'df -h': /dev/</match>" to reflect the new command.


ddpbsd:
I'm not sure exactly what happened here, but there is too much in the PR to go over.
PRs should be opened against MASTER first. That's where the new development happens. I'm not sure if stable is even a living branch at the moment.
Also, I'm not sure where the `df` change is in this diff, but `P` and `h` are incompatible together on at least OpenBSD.

PR #832, opened 2016-05-06
Add a rule to quiet 1002 from alerting on failed useradds.

ddpbsd:
Are useradd failures something we'd care about seeing? I have it set to level 0 at the moment, but it wouldn't be hard to convince me it should be higher.


ddpbsd:
I have no idea why the check is failing. I didn't set a decoder, and it doesn't find a decoder. So everything should be hunky dory.

PR #840, opened 2016-05-13
- GeoIP support in Alerts

atomicturtle:

This is a re-visit of dcids GeoIP patch, it adds the Geo Location of the IP address to the Alert and JSON output. It requires the GeoIP-devel package to build, and the GeoLiteCity.dat (included in GeoIP-GeoLite-data-extra package, or can be manually downloaded).

<global>
   <geoipdb>/usr/share/GeoIP/GeoLiteCity.dat</geoipdb>
</global>

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -19,7 +19,7 @@ OSSEC_USER_REM?=ossecr
 
 USE_PRELUDE?=no
 USE_ZEROMQ?=no
-USE_GEOIP?=no
+USE_GEOIP?=yes
 USE_INOTIFY=no
 
 ifneq (${TARGET},winagent)
--- a/src/Makefile
+++ b/src/Makefile
@@ -19,7 +19,7 @@ OSSEC_USER_REM?=ossecr
 
 USE_PRELUDE?=no
 USE_ZEROMQ?=no
-USE_GEOIP?=yes
+USE_GEOIP?=no
 USE_INOTIFY=no
 
 ifneq (${TARGET},winagent)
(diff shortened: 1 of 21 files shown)


atomicturtle:
This is probably going to break because it requires an external GeoIP library dependency

ddpbsd:
The GeoIP thing should be handled by Travis. I see the apt-get installation in the travis output.

aquerubin:
On Fri, 8 Jul 2016, Dan Parriott wrote:

> Merged #840.

Unfortunately this restores code that was/is IPv4 dependent in alert.c. 
It would be better to leave srcip and dstip as is and write them as 
strings (into the sql query string) instead of converting them to ints 
which would never work for IPv6 addresses.

Antonio Querubin
e-mail:  tony@lavanauts.org

ddpbsd:
@aquerubin Thanks for the report. I wanted to get this in so there would be plenty of time to fix any issues before 3.0. I'm going to open an issue on this so it doesn't get lost, and quote your post.

aquerubin:
On Fri, 8 Jul 2016, Dan Parriott wrote:

> @aquerubin Thanks for the report. I wanted to get this in so there would 
> be plenty of time to fix any issues before 3.0. I'm going to open an 
> issue on this so it doesn't get lost, and quote your post.

Should I submit the fix or do you want to do that?

Antonio Querubin
e-mail:  tony@lavanauts.org

ddpbsd:
@aquerubin Your fix would be much quicker and better than mine. I'd appreciate if you gave it a go!

PR #848, opened 2016-05-18
Fixed bug at logcollector that inhibited alerts about file reduction

vikman90:
When OSSEC detects that a file size has been reduced, it updates this size into a local variable, but else it doesn't update it, so Logcollector is able to detect once that the current size of the file is lower than when the file was opened.

This fix makes logcollector update the chached size at each verification, allowing it to be more accurate comparing sizes and to alert if the file was shrunk.

### DIFF
--- a/src/logcollector/logcollector.c
+++ b/src/logcollector/logcollector.c
@@ -481,10 +481,6 @@ void LogCollectorStart()
                     debug1("%s: DEBUG: File size reduced. %s",
                            ARGV0, logff[i].file);
 
-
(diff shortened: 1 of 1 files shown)


ddpbsd:
The windows agent build is seeing the following error, maybe just a missing include?
```
logcollector/logcollector.c: In function ‘LogCollectorStart’:
logcollector/logcollector.c:546:29: error: ‘lpFileInformation’ undeclared (first use in this function)
```

vikman90:
It was a variable scope problem. Fixed!

PR #856, opened 2016-06-01
Cherry picked fixes for 2.9

ddpbsd:
3 fixes for 2.9 that went in after the branch.


ddpbsd:
I've added a couple more things that would be nice to have in 2.9.

ddpbsd:
Added the Makefile change to not delete local_decoder.xml, even though that's not accepted yet for MASTER.

PR #862, opened 2016-06-08
Fixed some memory leaks at analysisd

vikman90:
These changes should fix some memory leaks that Valgrind warned me about on running analysisd.


atomicturtle:
It looks like this is segfaulting on ossec-logtest if the system is configured to using a rules_dir configuration like this:

<rule_dir pattern=".xml$">etc/rules.d</rule_dir>

PR #878, opened 2016-06-29
2.9rc2 - applying a couple of fixes

ddpbsd:
Apply the permission fix to allow unmerging of shared files. PR #870 
Chroot for logtest. PR #864


ddpbsd:
Added fixes to versioning.

Issue #877, opened 2016-06-28
Added two very important decoders for windows os


mfbiux:
Basically, the decoder named **windows** managed to decode events correctly but did not extract the source IP for failed auth attempts. This, in turn, cause the windows active response not to fire.

The extra rules are these:

<decoder name="windows-login">
  <type>windows</type>
  <prematch>^\d\d\d\d \w\w\w \d\d \d\d:\d\d:\d\d WinEvtLog: Security: .\* Source Network Address: \S+</prematch>
  <regex>^\d\d\d\d \w\w\w \d\d \d\d:\d\d:\d\d WinEvtLog: Security: (\w+)((\d+)): (.+): </regex>
  <regex>(.+): .+: (\S+): </regex>
  <regex>.+:.+:.+:.+:.+:.+:.+:.+:.+:.+:.+:.+:.+:.+:.+:.+:.+:.+:.+:.+:.+Source Network Address: (\d+.\d+.\d+.\d+)</regex>
  <order>status, id, extra_data, user, system_name, srcip</order>
  <fts>name, location, user, system_name</fts>
</decoder>

<decoder name="windows-sql-login">
  <type>windows</type>
  <prematch>^\d\d\d\d \w\w\w \d\d \d\d:\d\d:\d\d WinEvtLog: Application: AUDIT_FAILURE(._): MSSQL._ [CLIENT: \S+]</prematch>
  <regex>^\d\d\d\d \w\w\w \d\d \d\d:\d\d:\d\d WinEvtLog: Application: (\w+)((\d+)): (.+): </regex>
  <regex>(.+): .+: (\S+): </regex>
  <regex>.+[CLIENT: (\d+.\d+.\d+.\d+)]</regex>
  <order>status, id, extra_data, user, system_name, srcip</order>
  <fts>name, location, user, system_name</fts>
</decoder>

At the moment, they only work if used BEFORE the current **windows** rule. On my server installation, I moved both the windows rule and these two rules to the local_decoder and set the windows rule at the bottom.

They should be added to the main decoder.xml for future releases.

ddpbsd:
Can you provide log samples?

mfbiux:
This is a typical Windows event log with source IP. The default decoder will correctly identify it and decode it but it will not extract the **srcip** which makes it useless in terms of active responses. My decoder will both correctly decode and extract the **srcip**

[...]

PR #880, opened 2016-07-05
Change-up the Windows decoder a bit.

ddpbsd:
Break the Windows decoder down a bit. I see no regression in the 1 log sample I had, and even gained a potentially useful field (srcip). These decoder changes deserve some extensive testing though, to make sure I didn't break everything else.

I think this handles issue #877 and supersedes pull request #809


ddpbsd:
This is just getting silly, so I'm merging this. We had no one trying it out before it was merged. Now someone can complain if it's broken instead of waiting for reports that it's not.

jesuslinares:
There is something weird:
`<regex offset="after_parent">^\.+: (\w+)\(\d+\): (\.+): </regex>`
it should be:
`<regex offset="after_parent">^\.+: (\w+)\((\d+)\): (\.+): </regex>`

Both ways are working, but only the second one should work...

PR #885, opened 2016-07-07
Fix different representation of permisions display.

sleemburg:
Hi,

The agentless file permissions are stored in the 'database' as octal values.
The agentfull file permissions are stored in the 'database' as the decimal st_mode value.

When reporting the agentless changes in permissions, weird representations are shown.

In order not to disturb any existing databases or introduce conversions on both the code and the database, this patch just determines if the permissions in the 'database' are in octal or decimal st_mode repesentation and converts this correctly to a string.

Also included are:

Fix Makefile openssl commandline flag.
Fix fixed version in debian contrib.

And 'unset HISTFILE' before the linux ssh command in the expect script.

Kind regards,
Stephan

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -294,7 +294,7 @@ endif # USE_OPENSSL
 
 ifneq (,$(filter ${USE_OPENSSL},yes y Y 1))
 	DEFINES+=-DLIBOPENSSL_ENABLED
-	ifneq (${OPENSSL_LIBS},)
+	ifeq (${OPENSSL_LIBS},)
 		OSSEC_LDFLAGS+=-lssl -lcrypto
 	else
 		OSSEC_LDFLAGS+=${OPENSSL_LIBS}
--- a/contrib/debian-packages/ossec-hids/debian/patches/01_makefile.patch
+++ b/contrib/debian-packages/ossec-hids/debian/patches/01_makefile.patch
@@ -8,7 +8,7 @@ Index: ossec-hids-2.8.2/Makefile
 +# 06/15/2015
 +#
 +
-+DESTDIR=/
++DESTDIR?=/
 +DIR=$(DESTDIR)/var/ossec/
 +OSSEC_INIT=$(DIR)/etc/ossec-init.conf
 +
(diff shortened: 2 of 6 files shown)


ddpbsd:
It looks like this is failing in the winagent builds:
```
shared/read-agents.c: In function ‘agent_file_perm’:
shared/read-agents.c:1244:27: error: ‘S_ISUID’ undeclared (first use in this function)
shared/read-agents.c:1244:27: note: each undeclared identifier is reported only once for each function it appears in
shared/read-agents.c:1245:27: error: ‘S_IRGRP’ undeclared (first use in this function)
shared/read-agents.c:1246:27: error: ‘S_IWGRP’ undeclared (first use in this function)
shared/read-agents.c:1247:27: error: ‘S_ISGID’ undeclared (first use in this function)
shared/read-agents.c:1247:52: error: ‘S_IXGRP’ undeclared (first use in this function)
shared/read-agents.c:1248:27: error: ‘S_IROTH’ undeclared (first use in this function)
shared/read-agents.c:1249:27: error: ‘S_IWOTH’ undeclared (first use in this function)
shared/read-agents.c:1250:27: error: ‘S_ISVTX’ undeclared (first use in this function)
shared/read-agents.c:1250:52: error: ‘S_IXOTH’ undeclared (first use in this function)
make[1]: *** [shared/read-agents.o] Error 1
make[1]: *** Waiting for unfinished jobs....
```

sleemburg:
Hi I put it within ifndef WIN32, so windows compile should not pick it up.

src/Makefile
```diff
@@ -294,7 +294,7 @@ endif # USE_OPENSSL
 
 ifneq (,$(filter ${USE_OPENSSL},yes y Y 1))
 	DEFINES+=-DLIBOPENSSL_ENABLED
-	ifneq (${OPENSSL_LIBS},)
```
ddpbsd: I'm not convinced this is correct, but I could be wrong. It doesn't look like $OPENSSL_LIBS is set anywhere in the Makefile. so I'm guessing this means "if $OPENSSL_LIBS is NOT set, add the flags to LDFLAGS, but if it IS set, add $OPENSSL_LIBS instead to the LDFLAGS."

ddpbsd:
Just in case there is no notification for the code comment, I've added one about the Makefile change.

src/Makefile
sleemburg: Hi, sorry for the late reply.

What you are guessing is correct. The patch changes the makefile's semantics from:

if $OPENSSL_LIBS is NOT set, add $OPENSSL_LIBS to LDFLAGS else add fixed flags to LDFLAGS

into

if $OPENSSL_LIBS is NOT set, add fixed flags to LDFLAGS else add $OPENSSL_LIBS to LDFLAGS 

I have not checked if OPENSSL_LIBS could be set anywhere, I just fixed the bug.

Kind regards,
Stephan

ddpbsd: I think I get it now. I'll have to look up the ifeq/ifneq stuff. Thanks!

Issue #888, opened 2016-07-08
Pull Request #840 reverts some ipv6 support


ddpbsd:
From pull request #840 comment thread:

Reported by @aquerubin

> Unfortunately this restores code that was/is IPv4 dependent in alert.c.
> It would be better to leave srcip and dstip as is and write them as
> strings (into the sql query string) instead of converting them to ints
> which would never work for IPv6 addresses.

PR #891, opened 2016-07-12
dbd IPv6 fixes

aquerubin:
Revert 'NOT NULL' for src_ip and dst_ip. This was mistakenly added in commit c2a232b during the initial set of IPv6 patches (postgresql schema has already been fixed in commit cc83f1e).

Fixes issue #888.  Revert part of commit eed1722 as it breaks IPv6 support.  Continuing to use int fields to store IP addresses doesn't scale to IPv6.  Existing database schemas will need to be converted per the suggested release notes in ossec-docs issue #125.


aquerubin:
Obsoleted.

PR #893, opened 2016-07-12
Do not use absolute indexes in Prelude fields

antoineluong:
When defining Prelude fields with absolute indexes, the following error can occur:

> IDMEF: Cannot add object 'alert.target(0).file(0).File_Access(1).user_id.number': An index below the one specified is undefined.

This happens because alert.target(0).file(0).File_Access(0) has not been set previously.

Use the append (>>) method instead, supported since libprelude-0.9.10.

### DIFF
--- a/src/analysisd/output/prelude.c
+++ b/src/analysisd/output/prelude.c
@@ -27,9 +27,6 @@
 #define ANALYZER_MODEL "Ossec"
 #define ANALYZER_MANUFACTURER __site
 #define ANALYZER_VERSION __version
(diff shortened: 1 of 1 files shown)


ddpbsd:
How many people actually use the prelude support? I'd much rather drop it completely.

antoineluong:
Hard to say... I do know that Prelude is packaged in some Linux distributions, and that OSSEC is widely used when deploying Prelude as the recommended HIDS.

PR #894, opened 2016-07-12
Systemd services

ddpbsd:
Attempt to make the systemd service files load the services in the proper order. This is untested.


ddpbsd:
PR #895 contains these changes, and more.

PR #895, opened 2016-07-12
Systemd control

ddpbsd:
**Requires/includes PR #894**

This should allow ossec-control to utilize systemd to manage the ossec processes. Also untested, and kind of unsure if I finished it.

### DIFF
--- a/src/init/ossec-server.sh
+++ b/src/init/ossec-server.sh
@@ -109,10 +109,13 @@ enable()
 
     if [ "X$2" = "Xdatabase" ]; then
         echo "DB_DAEMON=ossec-dbd" >> ${PLIST};
+        touch ${DIR}/etc/.dbd
     elif [ "X$2" = "Xclient-syslog" ]; then
         echo "CSYSLOG_DAEMON=ossec-csyslogd" >> ${PLIST};
+        touch ${DIR}/etc/.csyslogd
     elif [ "X$2" = "Xagentless" ]; then
         echo "AGENTLESS_DAEMON=ossec-agentlessd" >> ${PLIST};
+        touch ${DIR}/etc/.agentlessd
     elif [ "X$2" = "Xdebug" ]; then
         echo "DEBUG_CLI=\"-d\"" >> ${PLIST};
     else
--- a/src/init/ossec-server.sh
+++ b/src/init/ossec-server.sh
@@ -140,10 +140,13 @@ disable()
 
     if [ "X$2" = "Xdatabase" ]; then
         echo "DB_DAEMON=\"\"" >> ${PLIST};
+        rm /var/ossec/etc/.dbd
     elif [ "X$2" = "Xclient-syslog" ]; then
         echo "CSYSLOG_DAEMON=\"\"" >> ${PLIST};
+        rm /var/ossec/etc/.csyslogd
     elif [ "X$2" = "Xagentless" ]; then
         echo "AGENTLESS_DAEMON=\"\"" >> ${PLIST};
+        rm /var/ossec/etc/.agentlessd
     elif [ "X$2" = "Xdebug" ]; then
         echo "DEBUG_CLI=\"\"" >> ${PLIST};
     else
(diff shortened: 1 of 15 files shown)


src/init/ossec-server.sh
```diff
@@ -192,6 +199,16 @@ start()
         echo "OSSEC analysisd: Testing rules failed. Configuration error. Exiting."
         exit 1;
     fi
+
+    ## If the system is Linux, look for systemctl. If that file exists, use it.
+    ## XXX - system paths and exact execution are probably wrong.
+    if [ X`uname` == "XLinux" ]; then
+        if [ -x /sbin/systemctl ]; then
+            /sbin/ssytemctl start ossec-server.service
```
jrossi: > ssytemctl 
Think this might need to be systemctl

ddpbsd:
Fixed the ssytemctl typo.

ddpbsd:
I'm lightly testing the agent bits on Ubuntu 16.whatever LTS. Seems to be going well-ish so far.

ddpbsd:
I think restart isn't working. Will test later and update.

src/systemd/agent/ossec-execd.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=OSSEC Execd
 PartOf=ossec-agent.target
+Before=ossec-ossec-syscheckd.service ossec-agentd.service ossec-logcollector.service
```
phamvuong: typo here ? should be ossec-syscheckd

src/init/ossec-server.sh
```diff
@@ -260,6 +283,14 @@ pstatus()
 stopa()
 {
     lock;
+
+    if [ X`uname` = "XLinux" ]; then
+        if [ -x ${SYSTEMCTL} ]; then
+            ${SYSTEMCTL} stop ossec-server.target
+        fi
+        exit 0
+    if
```
phamvuong: fi

src/systemd/agent/ossec-execd.service
ddpbsd: Thanks for spotting those mistakes. They should be corrected now.

phamvuong:
@ddpbsd I'm not sure if you want this kind of behavior: if one of the member services (such as execd, syscheckd, etc) is killed/crashed by some reason, the server/agent target should be stopped.
If this is the case then I think you may want to use this option in the target file
https://www.freedesktop.org/software/systemd/man/systemd.unit.html#BindsTo=

```
[Unit]
Description=OSSEC HIDS agent
After=network.target
BindsTo=ossec-execd.service ossec-syscheckd.service ossec-agentd.service ossec-logcollector.service
Requires=ossec-execd.service
Requires=ossec-syscheckd.service
Requires=ossec-agentd.service
Requires=ossec-logcollector.service

[Install]
WantedBy=multi-user.target
```

phamvuong:
For the reload option in ossec-control.sh, you can apply the same thing in systemd using below option https://www.freedesktop.org/software/systemd/man/systemd.unit.html#PropagatesReloadTo=

However, it is applied for systemd version 235 and later only.

src/systemd/server/ossec-syscheckd.service
```diff
@@ -1,6 +1,8 @@
 [Unit]
 Description=OSSEC syscheckd
 PartOf=ossec-server.target
+After=ossec-maild.service ossec-execd.service ossec-analysisd.service ossec-remoted.service
+Before=ossec-agentless.service ossec-csyslog.service ossec-dbd.service ossec-monitord.service ossec-remoted.service
```
phamvuong: ossec-remoted has been in "After" section

src/systemd/server/ossec-remoted.service
```diff
@@ -1,6 +1,8 @@
 [Unit]
 Description=OSSEC remoted
 PartOf=ossec-server.target
+After=ossec-maild.service ossec-execd.service ossec-analysisd.service ossec-logcollector
```
phamvuong: ossec-logcollector.service

Issue #900, opened 2016-07-18
Pull Request #862 segfaults with rules_dir


atomicturtle:
From Pull Request #862 comment thread:

"
It looks like this is segfaulting on ossec-logtest if the system is configured to use a rules_dir configuration like this:

<rule_dir pattern=".xml$">etc/rules.d</rule_dir>"

Blocking issue, We will need to resolve or otherwise revert this PR to release 2.9

vikman90:
I'm testing it but I can't reproduce the issue.

I moved `rules/ossec_rules.xml` to `etc/rules.d` and replaced `<include>ossec_rules.xml</include>` for `<rule_dir pattern=".xml$">etc/rules.d</rule_dir>` and `ossec-logtest` seems to work, both with the PR and without it.

Can you describe the steps you followed to get the segfault?

ddpbsd:
Possible fix: https://gist.github.com/ddpbsd/264813ddaeef22f632eb7f6ec0f7ba39

PR #909, opened 2016-07-29
- Bugfix from ddpbsd for the decoders.d/rules.d segfault in logtest. Fix for Issue #900

atomicturtle:


ddpbsd:
I haven't checked , does this need to be in 2.9?

PR #920, opened 2016-08-11
Bugfixes for OS_IPFound, OS_IPFoundList, OS_IsValidIP.

reyjrar:
OS_IsValidIP violated it's `const` constraint when the ip_addr element
was set to 'any'.  This caused some unexplained behavior when subsequent
calls.  This is corrected.

OS_IPFound and OS_IPFoundList where call sacmp(), but if OS_IsValidIP
was called with "any", the sockaddr_store.ss_family would be AF_INET6.
AFAICT, this meant any comparison of "any" to an IPv4 address would
fail.  This affected the ossec-remoted when validating IPv4 keys issued
by authd which set 'any' as the IP address to verify.

This would also cause failures *everywhere* in the rules, analysis,
exec, and activeresponse code where the src/dst IP being compared to any
was IPv4.

### DIFF
--- a/src/shared/validate_op.c
+++ b/src/shared/validate_op.c
@@ -153,6 +153,20 @@ int OS_IPFound(const char *ip_address, const os_ip *that_ip)
     int _true = 1;
     os_ip temp_ip;
 
+    /* If negate is set */
+    char *ip = that_ip->ip;
+    if (ip[0] == '!') {
+        ip++;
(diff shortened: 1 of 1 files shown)


reyjrar:
Testing this on my infrastructure to ensure it's the last bug in 2.9 that's breaking the OSSEC/authd stuff.

PR #931, opened 2016-08-25
fix common realloc mistake in custom_output_search_replace.c

bryongloden:
Passing one pointer into realloc() and assigning the result directly into that same pointer variable can cause a memory leak if the reallocation fails, because the original allocation will still exist. The correct way to do this is to use a temporary pointer variable.

Found by https://github.com/bryongloden/cppcheck


ddpbsd:
Thanks for the patch!

PR #934, opened 2016-08-30
Create OSSEC users and group as system members

vikman90:
This commit was originally uploaded to the Wazuh repository by @DLGolden:

>This change causes the Linux groupadd and useradd to flag the ossec group and users as system accounts. This is supported by every Linux system that I have researched and should be all; however, if for some reason the system account option is not supported, it will add the group and users exactly as it did prior to the change.

### DIFF
--- a/src/init/adduser.sh
+++ b/src/init/adduser.sh
@@ -49,8 +49,8 @@ else
 	# All current linux distributions should support system accounts for
 	# users/groups. If not, leave the GROUPADD/USERADD as it was before
 	# this change
-	sys_acct_chk () { 
-	    $1 --help | grep -e " *-r.*system account" >/dev/null 2>&1 && echo "$1 -r" || echo "$1"
+	sys_acct_chk () {
+	    $1 --help 2>&1 | grep -e " *-r.*system account" >/dev/null 2>&1 && echo "$1 -r" || echo "$1"
 	  }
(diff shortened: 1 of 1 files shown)


ddpbsd:
OpenBSD doesn't seem to like the change (group is still added, but this is ugly):
```
./init/adduser.sh ossec ossecm ossecr ossec /var/ossec
Wait for success...
groupadd: unknown option -- -
usage: groupadd [-ov] [-g gid] group
useradd: unknown option -- -
usage: useradd -D [-b base-directory] [-e expiry-time] [-f inactive-time]
               [-g gid | name | =uid] [-k skel-directory] [-L login-class]
               [-r low..high] [-s shell]
       useradd [-mov] [-b base-directory] [-c comment] [-d home-directory]
               [-e expiry-time] [-f inactive-time]
               [-G secondary-group[,group,...]] [-g gid | name | =uid]
               [-k skel-directory] [-L login-class] [-p password] [-r low..high]
               [-s shell] [-u uid] user
success
```

vikman90:
Thank you @ddpbsd for your comment.

That version of `useradd`/`groupadd` doesn't recognize `--help` and prints the help message on _stderr_. The new commit should fix this issue. I tested it on OpenBSD 5.9.

ddpbsd:
At a glance, it looks good, thanks!

PR #950, opened 2016-09-07
Exclude btrfs-Filesystem from searching for hidden files inside direc…

Joeri-MS:
…torys. (link count is always 1)


ddpbsd:
Thanks for the contribution!

PR #991, opened 2016-11-21
Change the default AR level to 7. 

ddpbsd:
The default email level is 7, and I'm not entirely convinced it's ok to block something automatically
with no notice whatsoever.
Brought up by Christina Plummer on the user list.

### DIFF
--- a/etc/ossec-local.conf
+++ b/etc/ossec-local.conf
@@ -146,7 +146,7 @@
       -->
     <command>host-deny</command>
     <location>local</location>
-    <level>6</level>
+    <level>7</level>
     <timeout>600</timeout>
   </active-response>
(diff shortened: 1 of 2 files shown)


atomicturtle:
I dont have a problem with bumping this, but I think the reasoning here was that it was an example of having a bucket for high-volume AR's that you want to block and ignore as a 6 and a more important block that you want a notification on at 7

ddpbsd:
That makes sense, buuuuuut I don't like blocking anything by default that isn't alerted on.
Either way, the documentation on this stuff needs to improve. I don't think anyone should expect AR to be flawless out of the box, but making it less surprising is a reasonable start.

PR #1001, opened 2016-12-19
Added 'auth failed' pattern to dovecot rule 9705

binrush:
On my server, dovecot `2.2.26.0` logs following messages on authentication failures:
```
Dec 19 06:21:06 ny dovecot: imap-login: Disconnected (auth failed, 7 attempts in 111 secs): user=<thousands>, method=PLAIN, rip=109.201.200.201, lip=67.205.141.203, sessi
on=<+hgd5vxDBMZtycjJ>
```

They are not covered by existing dovecot rules. My change aims to fix this.

### DIFF
--- a/etc/rules/dovecot_rules.xml
+++ b/etc/rules/dovecot_rules.xml
@@ -41,7 +41,7 @@
 
 <rule id="9705" level="5">
   <if_sid>9700</if_sid>
-  <match>user not found|User not known|unknown user</match>
+  <match>user not found|User not known|unknown user|auth failed</match>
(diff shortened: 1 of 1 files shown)


ddpbsd:
Thanks for the submission!

binrush:
@ddpbsd I made some additional research and it seems like the root problem is that current ossec rules/decoders are written for dovecot 1.2 and this little change does not provide complete support for 2.0. Is it possible to create separate rule/decoders sets to support dovecot 2.0 or we should just add new rules/decoders to existing groups?

Issue #1012, opened 2016-12-27
Handle "maximum authentication attempts exceeded"


leofeyer:
It seems that OSSEC does not yet handle the sshd "maximum authentication attempts exceeded" messsage:

```
Received From: ab->/var/log/auth.log
Rule: 1002 fired (level 2) -> "Unknown problem somewhere in the system."
Portion of the log(s):

Dec 27 03:23:51 r1 sshd[21183]: error: maximum authentication attempts exceeded for root from 183.106.179.x port 34100 ssh2 [preauth]
```

There should be a rule in the `sshd_rules.xml` file so this does not trigger a 1002, shouldn't it?

ddpbsd:
PR #1013 should handle this. Thanks for the sample!

leofeyer:
Although the fix is in the master branch (see https://github.com/ossec/ossec-hids/blob/master/etc/decoder.xml#L321-L326), it is not included in version 2.9.1 (see https://github.com/ossec/ossec-hids/blob/2.9.1/etc/decoder.xml). @ddpbsd Why is this?

leofeyer:
The same applies to the `sshd_rules.xml` file.

**Master branch**
https://github.com/ossec/ossec-hids/blob/master/etc/rules/sshd_rules.xml#L396-L401

**Version 2.9.1**
https://github.com/ossec/ossec-hids/blob/2.9.1/etc/rules/sshd_rules.xml

Rules no. 5754 to 5758 are missing in version 2.9.1.

ddpbsd:
Perhaps nobody backported them? If there was a pull request or issue asking for it, I missed them.

leofeyer:
Ok, so can I ask you to back port them then? 😄

atomicturtle:
Yeah I can do it for 2.9.2. Do you think you could go through the other rules to see if there are any other deviations? Thanks!

leofeyer:
Sure. The following files from the master branch all have rules which are not present in version 2.9.1:

* [dovecot_rules.xml](https://github.com/ossec/ossec-hids/blob/master/etc/rules/dovecot_rules.xml#L78-L88)
* [exim_rules.xml](https://github.com/ossec/ossec-hids/blob/master/etc/rules/exim_rules.xml) (does not exist in v2.9.1 at all)
* [msauth_rules.xml](https://github.com/ossec/ossec-hids/blob/master/etc/rules/msauth_rules.xml#L944-L968)
[...]

PR #1013, opened 2016-12-27
sshd support for maximum auth attempts logs

ddpbsd:
Add a rule, decoder, and test for max auth attempt messages

```
Dec 27 03:23:51 r1 sshd[21183]: error: maximum authentication attempts exceeded for root from 183.106.179.x port 34100 ssh2 [preauth]
```

This should complete issue #1012

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1013)

### DIFF
--- a/contrib/ossec-testing/tests/sshd.ini
+++ b/contrib/ossec-testing/tests/sshd.ini
@@ -139,3 +139,10 @@ rule = 5757
 alert = 0
 decoder = sshd
 
+[max auth attempts]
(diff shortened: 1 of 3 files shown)


leofeyer:
Great, thank you very much @ddpbsd.

leofeyer:
I think it should be `<decoder name="sshd-exceed">` instead of `ssh-exceed`, shouldn't it?

ddpbsd:
Thanks @leofeyer, I made the change.

PR #1014, opened 2016-12-28
Fix missing "fi" in ossec-control (server version) startup() script.

jbullard725:
The template that ultimately becomes ossec-control has a missing "fi".  It _seems_ that the only effect was throwing an error message (because it interpreted the "done" as closing the if block), but this should at least fix that.

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1014)

### DIFF
--- a/src/init/ossec-server.sh
+++ b/src/init/ossec-server.sh
@@ -219,6 +219,7 @@ start()
             else
                 echo "${i} already running..."
             fi
+        fi
     done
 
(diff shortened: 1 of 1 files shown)


jbullard725:
I will submit a different pull request...it turns out the closing "fi" was more misplaced than missing.  This change fixes the error but didn't solve the root problem of the maild never starting up...

PR #1016, opened 2016-12-28
Fix bug that prevents ossec-control from starting ossec-maild on server

jbullard725:
An "else" instead of "fi" (introduced in commit 89e7f5e)
was causing the the actual start up of ossec-maild to be skipped, even if the XML tag was
set to "yes".

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1016)

### DIFF
--- a/src/init/ossec-server.sh
+++ b/src/init/ossec-server.sh
@@ -204,21 +204,20 @@ start()
              if [ $? = 0 ]; then
                  continue
              fi
-        else
+        fi
 
-            pstatus ${i};
-            if [ $? = 0 ]; then
-                ${DIR}/bin/${i} ${DEBUG_CLI};
-                if [ $? != 0 ]; then
-                    echo "${i} did not start correctly.";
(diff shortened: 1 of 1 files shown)


src/init/ossec-server.sh
```diff
@@ -219,6 +219,7 @@ start()
             else
                 echo "${i} already running..."
             fi
+        fi
```
jbullard725: This change will be superseded by commit a48989506d8e1518a0bab5f313c4f283b94cb081 (sorry about that)

ddpbsd:
As soon as this is done testing, I'll meege it (or tomorrow, something like that). Thanks!

ddpbsd:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/ossec/ossec-hids/1016)*

PR #1017, opened 2016-12-29
Remoted canary violation

ddpbsd:
One of the free(ip_address) entries added in commit 0f776e8915be3421307b5c3885e05ef13011de17 causes remoted to crash with a canary violation.
```
OpenBSD 6.0-current (GENERIC.MP) #35: Wed Dec 14 10:38:03 MST 2016
[ddp@ix] :; ls -l /etc/malloc.conf
lrwxr-xr-x  1 root  wheel  2 Oct 31 11:09 /etc/malloc.conf -> SC
```

Actual error message:
```
ossec-remoted(55078) in free(): chunk canary corrupted 0x1bae51e6c910 0x4@0x4
```

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1017)

### DIFF
--- a/src/init/ossec-server.sh
+++ b/src/init/ossec-server.sh
@@ -214,7 +214,6 @@ start()
                 unlock;
                 exit 1;
             fi
-	fi
     done
 
     # After we start we give 2 seconds for the daemons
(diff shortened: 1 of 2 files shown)


ddpbsd:
I'm not sure what's different, but the `os_strdup` instead of strncpy seems to fix the issue as well. Hopefully that fixes it for the branches I didn't test.

PR #1020, opened 2017-01-04
Allow notify_timeout to be configured server-side.

friedkiwi:
It wasn't possible to configure the time-out on the server after which
to create a notification that a HIDS agent went down.

This patch defines a configuration file entry and allows the NOTIFY_TIME
setting to be configured for monitord.

See https://groups.google.com/forum/#!topic/ossec-list/VQHP8c-pYxA

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1020)


ddpbsd:
Thanks for the PR!

PR #1021, opened 2017-01-05
Fix Windows build

friedkiwi:
This PR fixes the following Windows-related build issues:

S_ISLNK usage
S_ISLNK isn't a supported call on Windows due to the lack of filesystem links like they exist on Unix-based systems. Other parts of the code properly check if WIN32 is defined, however, one piece was overlooked. This is fixed in one of the commits in this PR.

MINGW_HAS_SECURE_API Incorrectly Used
The MINGW_HAS_SECURE_API was used incorrectly and broke the Windows build - the correct usage is to define this to '1' instead of just defining it, as evident in the mingw headers.

Add Windows-specific output to .gitignore
Some of the build output files aren't added to the .gitignore, causing a potential source tree pollution when committing to the repository. Most notably, there was no exclusion for *.exe and *.dll files, while there was an exclusion for .so files. This discrepancy has been corrected.

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1021)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@
 *.a
 *.dSYM
 .DS_Store
+*.dll
+*.exe
 
 # Auto generated build files
 src/LOCATION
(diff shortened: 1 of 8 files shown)


ddpbsd:
Thanks a lot!

PR #1036, opened 2017-01-27
Add cpanel decoders and rules.

alex-front:
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1036)

### DIFF
--- a/etc/decoder.xml
+++ b/etc/decoder.xml
@@ -2936,8 +2936,7 @@ Jul 26 13:57:56 mx1.example.org outbound/smtp: 127.0.0.1 1406297159-06f4a35b4df2
 -->
 
 <decoder name="cpanel-access-failed">
-  <parent>web-accesslog</parent>
-  <prematch offset="after_parent">FAILED LOGIN</prematch>
+  <prematch>^\S+ \S+ \S+ [\d\d/\d\d/\d\d\d\d:\d\d:\d\d:\d\d \S*\d+] "\S+" FAILED LOGIN</prematch>
   <regex>^(\S+) \S+ (\S+)</regex>
   <order>srcip,user</order>
 </decoder>
--- a/etc/decoder.xml
+++ b/etc/decoder.xml
@@ -2942,15 +2942,15 @@ Jul 26 13:57:56 mx1.example.org outbound/smtp: 127.0.0.1 1406297159-06f4a35b4df2
   <prematch offset="after_parent">^SMTP call from </prematch>
   <regex offset="after_prematch">[(\S+)]:\d+ dropped: too many syntax or protocol errors</regex>
   <order>srcip</order>
+</decoder>
 
 
 <!-- cPanel decoder.
   - Examples:
-  - [2016-11-18 09:32:19 +0000] info [cpsrvd] 10.1.5.19 - admin "POST /login/?login_only=1 HTTP/1.1" FAILED LOGIN whostmgrd: user password hash is missing from system (user probably does not exist)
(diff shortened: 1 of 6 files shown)


etc/decoder.xml
```diff
@@ -2896,4 +2896,68 @@ Jul 26 13:57:56 mx1.example.org outbound/smtp: 127.0.0.1 1406297159-06f4a35b4df2
...
+  <regex>info [cpsrvd] (\S+) - (\S+)</regex>
+  <order>srcip,user</order>
+</decoder>
+
+<!-- cPanel decoder.
+  - Examples:
+  - 10.1.5.19 - paul [11/18/2016:09:35:43 -0000] "GET" FAILED LOGIN cpdavd: Could not fetch system home directory for paul
+-->
+
+<decoder name="cpanel-access-failed">
```
ddpbsd: It looks like this decoder is causing the failures, but I'm not sure why yet.

ddpbsd: The log message associated with this didn't seem to match a decoder (at least on my setup), so using this decoder things seem to work so far:
```
<decoder name="cpanel-access-failed">
  <!--<parent>web-accesslog</parent>
  <prematch offset="after_parent">^FAILED LOGIN</prematch>-->
  <prematch>^\S+ \S+ \S+ [\d\d/\d\d/\d\d\d\d:\d\d:\d\d:\d\d \S*\d+] "\S+" FAILED LOGIN</prematch>-->
  <regex>^(\S+) \S+ (\S+)</regex>
  <order>srcip,user</order>
</decoder>
```

alex-front: I have tested your proposed decoder and I am confirming it is working as expected and CI tests are passing. I have added the fix to pull request. Please merge.

alex-front: I think the postgresql_log log decoder should be fixed to avoid clashes with and duplication of the cpanel decoders and rules like I proposed.

ddpbsd: What's wrong with the postgresql_log decoder exactly?

etc/decoder.xml
```diff
@@ -2943,6 +2942,68 @@ Jul 26 13:57:56 mx1.example.org outbound/smtp: 127.0.0.1 1406297159-06f4a35b4df2
   <prematch offset="after_parent">^SMTP call from </prematch>
   <regex offset="after_prematch">[(\S+)]:\d+ dropped: too many syntax or protocol errors</regex>
   <order>srcip</order>
+
```
ddpbsd: You lost the `</decoder>` from the exim decoder here.

atomicturtle:
We should also capture these as tests in ossec-testing so we dont end up with collisions on other rules. As I recall cpanel environments frequently use windows timestamp format for example, I know that would overlap with the windows decoder

ddpbsd:
I want to hold off on these until PRs #1011  #1042  #1037  #1045  get merged. Those are a lot simpler, I think. then when that churn is done I will devote more time to this PR.

etc/decoder.xml
alex-front: Oops, fixed.

etc/decoder.xml
alex-front: The problem with postgresql_log is that it will decode messages like:
'[2017-02-03 01:21:31 -0500]' and not like '[2017-02-03 01:21:31 +0500]' (note the '+' sign). So to handle both cases I had to duplicate all rules and decoders for OSSEC. So one way to avoid this is to fix the postgresql_log decoder to be more specific to avoid it clashing with cpanel decoders altogether but I don't have much knowledge of their log format so can't propose how to do it.

PR #1037, opened 2017-01-27
exim rules and tests

ddpbsd:
This should fix the exim rules (tested with a fresh install).
Also create tests for exim and cpanel (hopefully rules will come later).

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1037)

### DIFF
--- a/etc/rules/syslog_rules.xml
+++ b/etc/rules/syslog_rules.xml
@@ -579,7 +579,7 @@
 <group name="syslog,dpkg,">
   <rule id="2900" level="0">
     <decoded_as>windows-date-format</decoded_as>
-    <regex>^\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d \w+ </regex>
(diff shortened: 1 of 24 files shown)


alex-front:
It was a good idea to fix the 2900 rule to avoid overriding it in the exim rules.

etc/rules/syslog_rules.xml
```diff
@@ -579,7 +579,7 @@
 <group name="syslog,dpkg,">
   <rule id="2900" level="0">
     <decoded_as>windows-date-format</decoded_as>
-    <regex>^\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d \w+ </regex>
+    <regex offet="after_parent">^startup |^status |^remove |^configure |^install |^purge |^trigproc </regex>
```
iasdeoupxe: Reported this "offet" typo as a new issue in https://github.com/ossec/ossec-hids/issues/1433

Issue #425, opened 2014-10-31
manage_agents unable to access /dev/random due to chroot


bitcomsecdev:
Hi,

OS: 14.10 (Utopic Unicorn)
Kernel: Linux 3.16.0-23-generic
OSSEC Version: snapshot from github an hour ago.

I installed OSSEC for the first time and was excited to get going. Once installed I tried to add an agent via manage_agents bin and got:

2014/10/31 09:34:55 manage_agents: ERROR: randombytes failed for all possible methods for accessing random data

Strace output:
[ output omitted for readability ]
setgroups(1, [1001])                    = 0
setresgid(-1, 1001, -1)                 = 0
setgid(1001)                            = 0
chdir("/var/ossec")                     = 0
chroot("/var/ossec")                    = 0
chdir("/")                              = 0

[ .. more omittions ..]

open("/dev/urandom", O_RDONLY)          = -1 ENOENT (No such file or directory)
open("/dev/random", O_RDONLY)           = -1 ENOENT (No such file or directory)
write(2, "2014/10/31 09:38:16 ", 202014/10/31 09:38:16 )    = 20
write(2, "manage_agents: ERROR: randombyte"..., 91manage_agents: ERROR: randombytes failed for all possible methods for accessing random data) = 91
write(2, "\n", 1
)                       = 1
exit_group(1)                           = ?
+++ exited with 1 +++

From what I gather manage_agents chroots itself, but doesn't take into account that it actually needs entropy from /dev/urandom and /dev/random.

If you need the entire strace log by all means just let me know.

pjmorr:
this issue 'seems' to be causing system wide effects??  I have a new server which is now no longer able to git clone or wget after doing the mount -o bind /dev dev/ from /var/ossec as root.
the error:
```
~# wget https://github.com/ossec/ossec-hids/archive/master.zip
Error in GnuTLS initialization: Failed to acquire random data.
ERROR: Failed to open cert /etc/ssl/certs/d7746a63.0: (-402).
ERROR: Failed to open cert /etc/ssl/certs/11f154d6.0: (-402).
ERROR: Failed to open cert /etc/ssl/certs/6e8bf996.0: (-402).
ERROR: Failed to open cert /etc/ssl/certs/9d6523ce.0: (-402).
ERROR: Failed to open cert /etc/ssl/certs/Global_Chambersign_Root_-_2008.pem: (-402).
ERROR: Failed to open cert /etc/ssl/certs/124bbd54.0: (-402).
[...]

PR #1040, opened 2017-01-31
Try to fix the manage_agents chroot (again)

ddpbsd:
Privsep_Chroot() chroots, so include that in the willchroot check.
Make sure the authfile always exists under the ossec dif (do not write to /etc/client.keys).

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1040)


ddpbsd:
Oops, forgot to mention that this possibly "completes" the issue #425 problems.

PR #1069, opened 2017-02-23
Add some wordpress rules 

ddpbsd:
from Icaro Torres at ibliss on the dev list.

*totally untested by me at this point.

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1069)


dcid:
These rules will generate a lot of false positives.

GET or POST requests to /wp-json/wp/v2/posts/ are not malicious per se as it is how the WordPress REST API works.

It should only be flagged if the id is set to [^0-9]. Note that the exploit also works against /wp-json/wp/v2/pages/ and any custom post type.

It is actually really tricky to detect it via OSSEC because the id can be pushed via the POST body, which is not logged or visible at that level.

ddpbsd:
Thanks for the comments @dcid 
I'll mark these as not ready yet, at least for now.

sempervictus:
An arachni or wpscan session with this in place might cause a log deluge without significant rate limiting. Every evasive client implements UA changes, and the advanced ones implement significant request encoding capabilities (so even if we couple this with modsec to capture the POST there's some work to be done after that).
It might be worthwhile to verify content of the subject post when one of these is detected to find any tags, includes, or "funky" content seeking to escape filters. Without subsequent validation of effect, its a probing action at most, and on a commonly matched regex, there are likely to be more FP logs than actionable intel on a WAN-facing host with significant traffic.

Issue #1084, opened 2017-03-07
*** glibc detected *** /var/ossec/bin/ossec-reportd


keith-mo:
Hi 

I'm trying create a couple of reports and i'm getting the following error since upgrading to 2.9.0

cat /var/ossec/logs/alerts/alerts.log | /var/ossec/bin/ossec-reportd -f group authentication_success 

2017/03/07 16:17:14 ossec-reportd: INFO: Started (pid: 31370).
*** glibc detected *** /var/ossec/bin/ossec-reportd: double free or corruption (out): 0x0000000001d9b990 ***
Aborted

PR #1088, opened 2017-03-09
Stop a double-free.

ddpbsd:
Reported by Keith-mo in issue #1084 
If this is the right fix, it'll have to be backported to 2.9.1.

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1088)

### DIFF
--- a/src/shared/report_op.c
+++ b/src/shared/report_op.c
@@ -522,7 +522,7 @@ void os_ReportdStart(report_filter *r_filter)
                     mgroup++;
                 }
 
-                free(mgroup);
+                //free(mgroup);
             } else {
                 tmp_str = al_data->group;
                 while (*tmp_str == ' ') {


src/shared/report_op.c
```diff
@@ -522,7 +522,7 @@ void os_ReportdStart(report_filter *r_filter)
                     mgroup++;
                 }
 
-                free(mgroup);
+                //free(mgroup);
```
jrossi: Why is this a double free?  This is a **mgroup is a list of pointers to list char's the functions initerates thought each member then frees them when done and finally at the free's the  array of arraies.

ddpbsd: I just used the terminology from the report. That free was making it crash, removing it stopped the crash. I can provide gdb output later.

zvanderbilt:
For whats it's worth. I manually applied this fix to the latest 2.9.2 release and reporting is working again without issue. I have yet to run an automated report but I expect that will complete as well.

hafeezy2j:
@ddpbsd in regards to issue#1227,  I have already applied the fix above before the upgrade. If reports are generated manually, they are working fine. However, automated/Scheduled reports are not working.

PR #1091, opened 2017-03-11
syscheck: whitelist for files in an sqlite database

ddpbsd:
Integrate a patch from Xavier Mertens to ignore MD5s for files in an sqlite database.
https://blog.rootshell.be/2013/05/13/improving-file-integrity-monitoring-with-ossec/

I have only done VERY light testing, so this needs to be tested before merge. For instance, I'm not sure the syscheck db entry gets updated if the file is in the sqlite db. Should it be? Not sure yet.

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1091)


src/Makefile
```diff
@@ -1049,11 +1048,11 @@ all_analysisd_libs += decoders.a decoders-test.a decoders-live.a
 
 
 analysisd/decoders/%-test.o: analysisd/decoders/%.c
-	${OSSEC_CC} ${OSSEC_CFLAGS} -DTESTRULE -DARGV0=\"ossec-analysisd\" -I./analysisd -I./analysisd/decoders -c $^ -o $@
+	${OSSEC_CC} ${OSSEC_CFLAGS} -DTESTRULE -DARGV0=\"ossec-analysisd\" -I./analysisd -I./analysisd/decoders -c $^ -lsqlite3 -o $@
```
jrossi: Why is the lib loaded on to each CC line?  Should the it not be appended to correct CFLAGS?

src/analysisd/decoders/syscheck.c
```diff
@@ -633,6 +674,40 @@ int DecodeSyscheck(Eventinfo *lf)
...
+     * 0:0:0:0:78f5c869675b1d09ddad870adad073f9:bd6c8d7a58b462aac86475e59af0e22954039c50
+     */
+    if (Config.md5db)  {
+        if ((p = extract_token(c_sum, ":", 4))) {
+            if (!validate_md5(p)) { /* Never trust input from other origin */
+                merror("%s: Not a valid MD5 hash: '%s'", ARGV0, p);
+                return(0);
+            }
+            debug1("%s: Checking MD5 '%s' in %s", ARGV0, p, Config.md5db);
+            if (!(error = sqlite3_open(Config.md5db, &conn))) {
```
jrossi: Could be completely wrong here, but open and closing the SQLite db for each operation seams incorrect.  SHould it not be done during once?

src/Makefile
ddpbsd: I had enough trouble figuring out how to get it working, that I didn't get around to "optimizing." I'll definitely work on this though, thanks!

src/analysisd/decoders/syscheck.c
ddpbsd: I pulled the code straight from Xavier, I haven't looked into it much more than that. Definitely something to work on though, thanks!

PR #1100, opened 2017-03-30
Open received files in binary mode cause of cr/lf and let hashes match.

gjocher:
We had issues with shared files on windows and tracked it down to a md5 hash missmatch caused by fopens textmode.
With this patch, received files are written in binary mode (and not deleted afterwards).

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1100)

### DIFF
--- a/src/client-agent/receiver-win.c
+++ b/src/client-agent/receiver-win.c
@@ -151,7 +151,7 @@ void *receiver_thread(__attribute__((unused)) void *none)
                              SHAREDCFG_DIR,
                              tmp_msg);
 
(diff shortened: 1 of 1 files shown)


nbuuck:
https://github.com/ossec/ossec-hids/pull/1207 was a duplicate of this but was already merged on 2017-07-24.

spaims:
Please can this be added to 2.9.4 #1350  
It is such a small change and has been missed in 2.9.2 and 2.9.3
Thanks

ddpbsd:
@spaims Done.

PR #1102, opened 2017-04-05
Fix TimeStamp

dvas0004:
The previous method (using ctime) has a number of issues that make it a bit weird to integrate with popular logging tools like ELK:

1. It appends a \n at the end of the timestamp which messes up date formatting for the unwary
2. It outputs time in the local timezone, without actually outputting the timezone... so logging tools like ELK will assume the date is in UTC, when actually it's GMT+2 or whatever... causing silly things like display logs in the future 

Easier method would be as proposed - simply output the timestamp in UNIX/EPOCH format. In this case it's millisecond epoch format - this is far more unambiguous for log parsers

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1102)

### DIFF
--- a/src/analysisd/format/to_json.c
+++ b/src/analysisd/format/to_json.c
@@ -18,11 +18,8 @@
 /* Convert Eventinfo to json */
 char *Eventinfo_to_jsonstr(const Eventinfo *lf)
 {
-    time_t c_time;
     char* time_string;
-    c_time = time(NULL);
-    time_string = ctime(&c_time);
-
+ 
     cJSON *root;
     cJSON *rule;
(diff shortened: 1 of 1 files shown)


ddpbsd:
I don't like changing things like this, but restricting it to the json log makes it a bit better.
Thanks for the PR!

dvas0004:
Thanks!

Best Regards,
David Vassallo

*--------------------------------*
*@dave_vassallo <https://twitter.com/dave_vassallo>*
*http://blog.davidvassallo.me <http://blog.davidvassallo.me>*

PR #1120, opened 2017-05-11
Pix decoders, Fixed 6-605004 and 6-605005

crlorentzen:
6-308001 is a very different format and cannot be in the same entry as the others
6-605004 and 6-605005 Also added support for action (Permitted, Denied), srcip, srcport, and User.

Tested with real traffic and with the ossec-logtest tool via this example log
%PIX-6-605004: Login denied from 192.168.1.100/25678 to inside:192.168.1.254/ssh for user "t@st123user" 

The original rule selects permitted or denied as the srcip.
       decoder: 'pix'
       id: '6-605004'
       srcip: 'denied'

Corrected rule produces this output
       decoder: 'pix'
       id: '6-605004'
       action: 'denied'
       srcip: '192.168.1.100'
       srcport: '25678'
       dstuser: 't@st123user'

This issue was found by our SOC when ELK was failing to import the json being produced by OSSEC-HIDS 2.9.0.

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1120)


ddpbsd:
Thanks! I'm going to start a `pix.ini` test using the provided log sample.

Issue #199, opened 2014-04-30
os_regex character maps


cgzones:
There are some inconsistencies in the regular expression character class maps:
- word (\w):
  - according to https://en.wikipedia.org/wiki/Regular_expression#Character_classes and http://msdn.microsoft.com/en-us/library/20bw873z%28v=vs.110%29.aspx#WordCharacter [a-zA-Z_0-9] should match
  - according to http://ossec-docs.readthedocs.org/en/latest/syntax/regex.html [a-zA-Z0-9] should match
  - and in code https://github.com/ossec/ossec-hids/blob/master/src/tests/test_os_regex.c#L494 [a-zA-Z_0-9@-] does match

I prefer [a-zA-Z_0-9] but this (migth) break existing decoders and rules so updating the docs to [a-zA-Z_0-9@-] is also reasonable.
- punctuation (\p):
  - according to http://ossec-docs.readthedocs.org/en/latest/syntax/regex.html ()*+,-.:;<=>?[] should match
  - in code ()*+,-.:;<=>?[] and additional !"'#$%&|{} match

I prefer the current code match.
- nonword (\W): 
  - here @ and - does not count as word but 127 (DEL)

I prefer exact inverse of word.

Question:
- space (\s) / tab (\t)
  - is it reasonable to have two regular expressions for this?
    According to http://msdn.microsoft.com/en-us/library/20bw873z%28v=vs.110%29.aspx#WhitespaceCharacter \s should match \t also. But \s is correct documented here: http://ossec-docs.readthedocs.org/en/latest/syntax/regex.html

PR #1124, opened 2017-05-15
Fixes for PIX 6-308001 6-605005, and 6-605004

crlorentzen:
Through further testing with PIX and ASA we have updated the decoder once more to support the current state of the ASA software. 

It was noticed that in Cisco ASA as apposed to PIX. the username is not provided in 6-605004. We think this is a security enhancement Cisco made, if a user fails to log in it is possible that a password was improperly put into the username field and thus it shouldn't be logged. Cisco instead puts asterisks(*) in place for the user on 605004

The only way we could see to fix the decoder for this was to remove the username from being selected. Perhaps someone with a better understanding of OSSEC regex could create something better.

Some more example logs and their phase 2 output before the fix.
Examples are also in the comment section of the decoder.xml.
```
%ASA-6-308001: Console enable password incorrect for 3 tries (from ssh (remote 198.18.1.100))
       decoder: 'pix'
       id: '6-308001'
       srcip: 'enable'


%ASA-6-605005: Login permitted from 198.18.1.100/47849 to outside:198.18.1.254/ssh for user "t@stuser123"
       decoder: 'pix'
       id: '6-605004'
       action: 'denied'
       srcip: '198.18.1.100'
       srcport: '56332'
       dstuser: 't@stuser123'

%ASA-6-605004: Login denied from 198.18.1.100/56332 to outside:198.18.1.254/ssh for user "*****"

       decoder: 'pix'
```

Log test output after the fix.
```
%ASA-6-308001: Console enable password incorrect for 3 tries (from ssh (remote 198.18.1.100))
       decoder: 'pix'
       id: '6-308001'

%ASA-6-605005: Login permitted from 198.18.1.100/47849 to outside:198.18.1.254/ssh for user "us3rn@m3"
       decoder: 'pix'
       id: '6-605005'
       action: 'permitted'
       srcip: '198.18.1.100'
       srcport: '47849'
       dstuser: 'us3rn@m3'

%ASA-6-605004: Login denied from 198.18.1.100/56332 to outside:198.18.1.254/ssh for user "*****" 
       decoder: 'pix'
       id: '6-605004'
       action: 'denied'
       srcip: '198.18.1.100'
       srcport: '56332'
```

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1124)

### DIFF
--- a/etc/decoder.xml
+++ b/etc/decoder.xml
@@ -1347,10 +1347,10 @@ Jan  8 19:32:41 tp.lan dropbear[15165]: Pubkey auth succeeded for 'root' with ke
   - %PIX-6-305012: Teardown dynamic UDP translation from inside:1.1.1.1/12 to outside:1.2.1.2/11 duration 0:00:11.
   - %PIX-3-305005: No translation group found for icmp src outside:x.x.x.x dst inside:x.x.x.x (type 3, code 0)
   - %ASA-2-106001: Inbound TCP connection denied from 1.2.3.4/1234 to 213.207.99.248/445 flags SYN on interface outside (Message repeated 2 times)
-	- %ASA-6-308001: Console enable password incorrect for 3 tries (from ssh (remote 198.18.1.100))
+  - %ASA-6-308001: Console enable password incorrect for 3 tries (from ssh (remote 198.18.1.100))
   - %PIX-6-605004: Login denied from 192.168.2.10/32597 to outside:192.168.2.14/ssh for user "root"
   - %ASA-6-605004: Login denied from 198.18.1.100/56332 to outside:198.18.1.254/ssh for user "*****" 
-	- %ASA-6-605005: Login permitted from 198.18.1.100/47849 to outside:198.18.1.254/ssh for user "us3rn@m3"
+  - %ASA-6-605005: Login permitted from 198.18.1.100/47849 to outside:198.18.1.254/ssh for user "us3rn@m3"
   - %PIX-6-605005: Login permitted from 192.168.1.2/2953 to inside:192.168.1.1/telnet for user ""
   - %PIX-6-305011: Built dynamic UDP translation from inside:192.168.1.2/1026 to outside:192.168.2.14/1163
   - %PIX-6-305011: Built dynamic TCP translation from inside:192.168.1.3/54946 to outside:192.168.2.14/1033
--- a/etc/decoder.xml
+++ b/etc/decoder.xml
@@ -1457,19 +1457,26 @@ Jan  8 19:32:41 tp.lan dropbear[15165]: Pubkey auth succeeded for 'root' with ke
   <order>id, srcip</order>
 </decoder>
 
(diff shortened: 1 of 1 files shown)


crlorentzen:
If someone can help me figure out how to accept either word characters, or asterisks (*) between the quotes (") then I'd be happy to update the single rule to include username in all cases.

ddpbsd:
`\S` should allow any non-whitespace characters.

crlorentzen:
While that is true, it unfortunately then chomps the " characters and fails to match properly.

ddpbsd:
I've run into this before. Can't remember how I solved it off hand.

Know1:
Not sure if you ever figured it out but this piece of Regex maybe what you need 
works for these test cases
  "*****"
  "us3rn@m3"
  "Testuser123"

`\"[aA-zZ|0-9|\*|\@]+\"`

Know1:
According to regex101 this snippet is faster, but based upon what you said in earlier comments may not work
`\"(\S+|\*+)\"`

Know1:
According to regex101 this snippet maybe faster depending on lang and other factors
Based upon comments above it may also not work correctly so YMMV
`\"(\S+|\*+)\"`

crlorentzen:
@Know1 unfortunately OSSEC does not use perl compatible regex (PCRE) https://github.com/ossec/ossec-hids/issues/199 So while a regex tester may indicate success, the only true test for OSSEC is to use the ossec-logtest tool which comes with the installation. I may have missued the term chomp, but the \S includes the quote (") character and doesn't work. Also OSSEC regex does not appear to support creating classes (e.g. [aA-zZ0-9*@]), although I haven't retested this in awhile, and my role on the ossec project at work is over, so I may not have time to retest.

Jupnnn:
> Actually after further thinking about \S matches all non space so should work. Greediness should exclude the chomping of " I plugged \S only into a regex tester with the above full strings and the snippet matched so I've gone back to the drawing board and come up with this that I tested against the following full strings
> 
> %ASA-6-605005: Login permitted from 198.18.1.100/47849 to outside:198.18.1.254/ssh for user "us3rn@m3" %ASA-6-605005: Login permitted from 198.18.1.100/47849 to outside:198.18.1.254/ssh for user "t@stuser123" %ASA-6-605004: Login denied from 198.18.1.100/56332 to outside:198.18.1.254/ssh for user "*****"
> 
> This piece of code breaks them out into $1, $2, $3, $4, $5 `<regex offset="after_parent">^(\S+): Login (\S+) from (\S+)\/(\d+).*user \"(\S+)\"</regex>`

Hello Sunshine

atomicturtle:
Superseded by https://github.com/ossec/ossec-hids/compare/main...atomicturtle:fix/1124-pix-asa-login-decoders

Thanks @crlorentzen — revisited with PCRE2 so 605004/605005 keep id/srcip for
masked/empty users (and @ usernames), and 308001 extracts a clean remote IP.

PR #1158, opened 2017-06-29
Update "shell" on ossec-hids-solaris.init

pedroflor:
"sh" shell generates error, but using "bash" it gets solved.
Tested on Solaris 11.2 Intel & SPARC

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1158)


ddpbsd:
Is bash installed by default on newer Solaris versions? What errors does `/bin/sh` give?

PR #1159, opened 2017-06-29
Update Makefile for Solaris installation

pedroflor:
This hack fixes the error generated on the population of the directory zoneinfo.
Maybe a better solution is needed.
Tested on Solaris 11.2 Intel & SPARC

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1159)


ddpbsd:
Is `install` not installed by default on Solaris?

ddpbsd:
Just looking back over this, and the change doesn't take owner and group into account.

PR #1168, opened 2017-06-30
Fixed typo correction on the auth help menu.

hcw2016:
Typo fixed in the auth menu.

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1168)

### DIFF
--- a/ossec-hids
+++ b/ossec-hids
@@ -1,0 +1,1444 @@
+
+
+
+
+
+
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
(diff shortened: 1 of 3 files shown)


ossec-hids
```diff
@@ -0,0 +1,1444 @@
+
```
ddpbsd: I think this file was mistakenly added. You should be able to remove it with `git rm ossec-hids && git commit`

Then a `git push` should remove it from the pull request.

ddpbsd:
If the `ossec-hids` file is removed, this can be accepted. Thanks!

hcw2016:
I pushed it back to you. Let me know if there is anything else!


Thanks,

Cody

PR #1170, opened 2017-06-30
These are patches to various src files for winagents.

AnthonyLoganMitchell:
These are different patches for getting the "makefile" to compile for winagents.

---
This change is [Reviewable](https://reviewable.io/reviews/ossec/ossec-hids/1170)


weisst:
@AnthonyLoganMitchell  i open administrator cmd and run agent-auth.exe with args, but it's crash,crash at os_net.c 
`s = getaddrinfo(_ip, _port, &hints, &result);`

ddpbsd:
The `index.html` files seem to have been added by accident.

AnthonyLoganMitchell:
ok so the added index files and the agent-auth.exe should now be working properly.

PR #1179, opened 2017-07-07
Fix README display IP as string

hugoboos:


aquerubin:
Aren't these stored as strings in the DB now?

Tony

ddpbsd:
From the mysql.schema:
```
    src_ip          VARCHAR(46),
    dst_ip          VARCHAR(46),
```

aquerubin:
Right.  So inet_ntoa in the select example isn't required any more is it?

Tony

hugoboos:
I've changed the column types to `VARCHAR(46)` but still a number is stored for the IP address.

aquerubin:
Were the numbers stored in the DB before the upgrade to 2.9?  If they were already in the DB then you’ll need to convert the existing data and schema using the script in src/os_dbd/convert-dbd-ipv6.sql.  Instructions are embedded in the script.  If you already converted the DB schema but the addresses are still being stored as numbers then we’ve got a separate problem.

PR #1200, opened 2017-07-15
Fixes newer (?) versions of check.

ddpbsd:
Some versions of check require `-lsubunit`
Hopefully this doesn't affect the version on travis.


ddpbsd:
Doesn't work with travis. Back to the drawing board.

PR #1209, opened 2017-07-27
Rule update

atomicturtle:


ddpbsd:
It looks like you pulled the github webpage for the `sshd_rules.xml` file instead of the raw content. I do that all the time.

PR #1210, opened 2017-07-28
Allow authd to require a password for agents to get a key

ddpbsd:
Based on the pull requests from @hcw2016 which I think were based on the Wazuh fork.

I removed chrooting from authd, which I'm not entirely happy with. But the `/dev/urandom` issue popped up again. `ossec-authd` might be something we can just require someone to create the device though?


ddpbsd:
This of course is built on the zlib update.

PR #1235, opened 2017-08-29
Spelling

jsoref:
fwiw, this is a subset of wazuh/wazuh#171


ddpbsd:
I don't like changing `groupping` at the moment, for backwards compatibility reasons. Everything else I saw looked good though

jsoref:
Dropped

atomicturtle:
Ha, this is fantastic. I wonder how many of those were my fault :P

jsoref:
fwiw, this is more or less about average for spelling errors in a project. (I haven't actually generated a metric for this, but someone could easily build one by reviewing the PRs I've made against dozens of projects...)

ddpbsd:
Thanks for going through all of those!

PR #1243, opened 2017-09-01
Better support for custom CFLAGS/CXXFLAGS/LDFLAGS

### LINKED ISSUE #1237: Support custom CFLAGS/CXXFLAGS/LDFLAGS
The HardenedBSD ports tree by default attempts to compile every application with PIE. This means that we pass in via the environment custom CFLAGS, CXXFLAGS, and LDFLAGS:

```
CFLAGS+= -fPIC -fPIE
CXXFLAGS+= -fPIC -fPIE
LDFLAGS+= -pie
```

This allows applications to take full advantage of ASLR. I'm having trouble making out how the ossec build process works, but it seems to me that it doesn't obey custom CFLAGS/CXXFLAGS/LDFLAGS.

ddpbsd:
Resolves #1237 
lua did not respect the user's custom variables, it uses MYCFLAGS and friends instead.
So we can set those to the appropriate variable.
Still needs more testing to make sure nothing breaks too much.


jrossi:
My guess is that you did not want the asterisks changes included in the PR?

ddpbsd:
Quite right, thanks @jrossi 
Closing this and opening a new one so I don't have to go through removing the asterisk stuff :-D

PR #1244, opened 2017-09-02
Trying to make the code more readable

flaviommedeiros:
Hi all,

I'm running some C code analysis tool that pointed me to this small code fragment. I'm suggesting this patch to try to make the source code more understandable.

Would you guys agree with this type of changes? The tools pointed me to almost hundreds cases like this one in the source code of the project.


ddpbsd:
No response from @flaviommedeiros and the code doesn't bother me.

Issue #1145, opened 2017-06-16
2.9 - networking broken on RHEL7.3 due to getaddrinfo/ipv6 changes


dwendt:
Upgrading an RHEL environment to 2.9, this definition breaks getaddrinfo. 

https://github.com/ossec/ossec-hids/blob/ef46a72250c84c068751a0161d8b5fb009fb7699/src/os_net/os_net.h#L22

It makes the addrinfo struct on line 52 have some options that make getaddrinfo fail. When the macro is undefined, it uses the other options which work perfectly fine. Not sure what those options are for (ipv6?) to begin with or I'd provide a patch suggestion....

```
Red Hat Enterprise Linux Server release 7.3 (Maipo)
Linux xxxxxxxxxxx 3.10.0-514.16.1.el7.x86_64 #1 SMP Wed Apr 12 10:03:19 PDT 2017 x86_64 x86_64 x86_64 GNU/Linux
```

ddpbsd:
Did you remove/disable ipv6 support?

mkvocka:
If the question was to me - yes I did. OSSEC syscheckd is commonly used in environments where PCI DSS compliance must be met. PCI DSS also requires disabling unused services, IPv6 in this case.

dwendt:
@ddpbsd welp, yep, ipv6 was disabled.
```
# sysctl -p
net.ipv6.conf.all.disable_ipv6 = 1
```

what would a good fix for this look like? if getaddrinfo fails for AF_INET6, attempt again with AF_INET?

ddpbsd:
Does PCI DSS actually define IPv6 as a service? That's crazy.
If anyone has a suggestion on how to change things so that these niche systems can be better supported without hampering modern systems, I'm all ears.

aquerubin:
Not specifically but it does say disable unused services.  Security admins have been interpreting that to mean disable IPv6 if it hasn't been deployed.  DSS also mandates the use of NAT.  That's a non-starter if IPv6 is in use.  The DSS has not really kept up with the times.

I've been wondering that if instead of breaking the network stack by disabling IPv6 completely, just disable it on individual physical interfaces only.  That reduces the attack surface until IPv6 is actually deployed.  I've not had a chance to test how well this would work in practice.

Tony

ddpbsd:
And that would make more sense to me, as well as possibly blocking IPv6 at the host's firewall.

robthralls:
[...]

PR #1259, opened 2017-09-21
if getaddrinfo fails with EAI_FAMILY try ipv4

ddpbsd:
Attempt to fix issue #1145 by possibly working with ipv4 only hosts.
Testing would be nice, I'm unable to do much right now.


ddpbsd:
As was pointed out, EAI_NONAME is returned, not EAI_FAMILY. The diff accounts for both.

PR #1260, opened 2017-09-24
Update ownCloud decoder for oC 10

kdslkdsaldsal:
This is a follow-up of https://github.com/ossec/ossec-hids/pull/1246

Seems they have changed the log order / syntax once more in ownCloud 10: https://github.com/owncloud/core/pull/27562


atomicturtle:
Thanks again for the PR!

PR #1291, opened 2017-10-21
New known malicious user agents

kdslkdsaldsal:
See e.g. http://www.skepticism.us/2015/05/new-malware-user-agent-value-jorgee/ (might be a good source for a few more malicious user agents)

Cross-Ref: https://github.com/wazuh/wazuh-ruleset/pull/79

Log-Sampe is:

``
1.2.3.4 - - [21/Oct/2017:16:27:00 +0200] "HEAD http://4.3.2.1:80/phppma/ HTTP/1.1" 404 0 "-" "Mozilla/5.0 Jorgee"
``

### DIFF
--- a/etc/rules/web_appsec_rules.xml
+++ b/etc/rules/web_appsec_rules.xml
@@ -88,7 +88,7 @@
   <!-- BAD/Annoying user agents -->
   <rule id="31508" level="6">
     <if_sid>31100</if_sid>
-    <match> "ZmEu"| "libwww-perl/|"the beast"|"Morfeus|"ZmEu|"Nikto|"w3af.sourceforge.net|MJ12bot/v</match>
+    <match> "ZmEu"| "libwww-perl/|"the beast"|"Morfeus|"ZmEu|"Nikto|"w3af.sourceforge.net|MJ12bot/v| Jorgee"|"Proxy Gear Pro|"DataCha0s</match>
     <description>Blacklisted user agent (known malicious user agent).</description>
   </rule>
 


kdslkdsaldsal:
Also added http://www.skepticism.us/2015/06/another-user-agent-to-blacklist-proxy-gear-pro2-1/ and another one.

Not quite sure if a few of the others or stuff like the following:

> Or they do something stupid like surround the user-agent string with double-quotes (see the first RewriteCond statement below).

from http://www.skepticism.us/2015/04/rejecting-http-requests-from-bad-actors/ should be added too?

kdslkdsaldsal:
Adding to this i'm also see quite a lot of malicious / probing requests from the following User-Agent:

```
1.2.3.4 - - [20/Oct/2017:23:21:48 +0200] "GET /mySqlDumper HTTP/1.1" 404 162 "-" "Python-urllib/2.7"
```

but i guess the Python-urllib is a too generic User-Agent to add here?

ddpbsd:
I think python's urllib would provide way too many false positives. I know I use it frequently enough to be annoying. Thanks for the updates though, they are appreciated!

kdslkdsaldsal:
Hey, thanks for your feedback. I thought about the same for Python-urllib 😃 

Besides the mentioned doubled quotes above sometimes i'm also seeing something like:

``
"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/53.0.2785.143 Safari/537.36"
``

Looks also quite suspicious if Chrome and Safari are mentioned in the same User-Agent string.

PR #1297, opened 2017-10-27
Named rules improvement

ncaio:
Protection against a host that don't has recursion permission and/or authorization to local cache query. there is a different between allow-query and allow-query-cache. 

ref: 
* https://kb.isc.org/article/AA-00503/0/Whats-the-difference-between-allow-query-cache-and-allow-recursion.html
* http://www.zytrax.com/books/dns/ch7/queries.html#allow-query-cache

### DIFF
--- a/etc/rules/named_rules.xml
+++ b/etc/rules/named_rules.xml
@@ -68,11 +68,12 @@
     <description>DNS update using RFC2136 Dynamic protocol.</description>
   </rule>
 
-  <rule id="12108" level="0">
+  <rule id="12108" level="5">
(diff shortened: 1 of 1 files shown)


ddpbsd:
Thanks for the pull request. [The tests](https://github.com/ossec/ossec-hids/blob/master/contrib/ossec-testing/tests/named.ini) will have to be adjusted for the new alert level. After that, I think this can be merged.

ddpbsd:
New PR #1335 covers this PR with the tests fix.

PR #1302, opened 2017-10-30
OSSEC Pagerduty Script

johnsiciliano:
This utilizes the pagerduty API to talk to pagerduty via OSSEC active-response


calve:
imho `$postfile` should be deleted after execution

PR #1304, opened 2017-10-31
Don't display the errno, show the error message.

ddpbsd:
I don't want to have to look up errno 28 because I'm lazy.


atomicturtle:
You know thats right

Issue #1065, opened 2017-02-17
Problem with negated Cidr in ossec-2.9


Azidburn:
Hey, i updated to ossec 2.9.0 and got errors when i tested the rules.

> *** Error in `bin/ossec-logtest': free(): invalid pointer: 0x0000000001c34031 ***

I searched which rule makes the problem. The problem caused by negation of the cidr in the following rule.

```
  <rule id="73126" level="7">
    <dstip>!xxx.xxx.xxx.0/23</dstip>
    <match> S </match>
    <description>Description</description>
  </rule>
```

Chrisfiveash8:
HEY FUCKING MAKE THIS SAMSUNG WORK! ! I need to bypass that google lock
screen. Tell me how I've always wanted a Samsung and it is really getting
on my nerves that you sit there and watch me suck on this thing for hours
and don't just give me the solution I need the answer now I'm not playing
anymore

ddpbsd:
I can reproduce this:
```
(gdb) run -qt
Starting program: /var/ossec/bin/ossec-logtest -qt
2017/02/17 13:58:47 ossec-testrule: INFO: Reading the lists file: 'rules/lists/ossec.block'
ossec-logtest(68968) in free(): chunk canary corrupted 0x1468f59d6631 0xf@0xe

Program received signal SIGABRT, Aborted.
0x000014693971e5ea in thrkill () at <stdin>:2
2       <stdin>: No such file or directory.
        in <stdin>
Current language:  auto; currently asm
(gdb) bt
#0  0x000014693971e5ea in thrkill () at <stdin>:2
#1  0x000014693976bd89 in *_libc_abort () at /usr/src/lib/libc/stdlib/abort.c:52
#2  0x000014693971a412 in wrterror (d=Variable "d" is not available.
) at /usr/src/lib/libc/stdlib/malloc.c:303
#3  0x000014693971a4e2 in validate_canary (d=Variable "d" is not available.
) at /usr/src/lib/libc/stdlib/malloc.c:1042
#4  0x000014693971a593 in find_chunknum (d=0x14693c02aa10, r=Variable "r" is not available.
) at /usr/src/lib/libc/stdlib/malloc.c:1061
#5  0x000014693971b386 in ofree (argpool=0x14693c02aa10, p=0x1468f59d6631) at /usr/src/lib/libc/stdlib/malloc.c:1359
[...]

PR #1333, opened 2017-12-19
Make negating a cidr in the rules work again

ddpbsd:
Fix issue #1065. Fix from @calve 

This includes the reports changes I made, so PR #1330  needs to go in first (after I fix it I guess)

### DIFF
--- a/src/monitord/sendcustomemail.c
+++ b/src/monitord/sendcustomemail.c
@@ -1,0 +1,314 @@
+/* Copyright (C) 2009 Trend Micro Inc.
+ * All rights reserved.
+ *
+ * This program is a free software; you can redistribute it
+ * and/or modify it under the terms of the GNU General Public
+ * License (version 2) as published by the FSF - Free Software
+ * Foundation
+ */
+
+/* Basic e-mailing operations */
+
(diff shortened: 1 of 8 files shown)


src/monitord/sendcustomemail.c
```diff
@@ -0,0 +1,314 @@
...
+#include "os_net/os_net.h"
+
+/* Return codes (from SMTP server) */
+#define VALIDBANNER     "220"
+#define VALIDMAIL       "250"
+#define VALIDDATA       "354"
+
+/* Default values used to connect */
+#define SMTP_DEFAULT_PORT   "25"
+#define HELOMSG             "Helo notify.ossec.net\r\n"
```
calve: hi,

no idea what you trying to do here, but I would like to bring your attention on that hardcoded `ossec.net` value. It might lead to

0. data leak from ossec users to notify.ossec.net
0. dos on notify.ossec.net

If this is the expected behaviour, maybe add a comment describing the goal of this file ?

ddpbsd: That was unrelated to the intent of this pull request. I've closed this one to cut down on the confusion.

PR #1348, opened 2018-01-03
Add SQLite3 syscheck database support to analysisd

ddpbsd:
**WARNING:** Very experimental!! (You can tell I'm serious because I used more than one exclamation point!)

Set `<database_type>sqlite</database_type>` in the `<global>` section of the `ossec.conf` to enable this.
The database will be in `/var/ossec/queue/syscheck/syscheck.db` and the schema has to be manually installed:
`sqlite3 /var/ossec/queue/syscheck/syscheck.db < src/analysisd/db/syscheck.schema`

This should change in the future if we decide to go forward with this.

Some caveats:

- Tested on a small agent/server setup. I don't have a lot of agents to pound sqlite, so it could be horrible with a decent number of agents.
- I think it will silently fail if the schema isn't loaded. This should all be automated at some point, but I haven't gotten there yet.
- The tools do not know how to work with this. `syscheck_control` and `syscheck_update` will need to be updated eventually.
- Documentation will need to be updated
- Probably more.


ddpbsd:
The more I think a about this, the less I like it. It might be ok for single systems, but not a server with a number of agents. I might be able to change it up a bit to make other databases an option.

reyjrar:
I was thinking of porting this code backwards in to the agent's syscheckd itself.  This way, locally running configuration management tools like Ansible, Chef, or Puppet could update their local DB's.  I think it's useful to still report the file changes to the server, but at a lower level.

I'm uncertain how this will work in the cases of an inotify-based monitoring as the file change may be detected before the configuration management engine has completed it's housekeeping and updated the checksums in the db.  I'm not familiar with the inotify API, so I'll have to see what info is available in those cases and maybe delay the agent's look up/reporting of the event.

PR #1360, opened 2018-02-01
Allow TLS Email sends as a compile-time option

alexbartlow:
To avoid setting up and configuring a sendmail server as a proxy (https://www.digitalocean.com/community/tutorials/how-to-send-email-through-an-external-smtp-service-with-sendmail-on-freebsd-10-1) We'd like the ability to simply point ossec-hids at an existing SMTP server.

Instead of re-implementing the wheel, I've added a compile-time flag to use curl to send with credentials over TLS to an external SMTP server. When compiling ossec-hids, you can use the `SENDMAIL_CURL=1` env var to trigger a build that uses curl to send emails to the server you specify, instead of passing the information to a local sendmail install.

We've been using this patch ourselves, and it's working great.

I thought I'd at least offer it up to the community, though I can see the virtue of not wanting to take on this added complexity concern within the product itself. It does, however, greatly simplify the process of setting it up. Here's a snip from our relevant chef cookbook:

```
  #config template
  <global>
    <email_notification>yes</email_notification>
    <email_to><%= @smtp_to %></email_to>
    <smtp_server><%= @smtp_server %></smtp_server>
    <email_from><%= @smtp_from %></email_from>
    <auth_smtp>yes</auth_smtp>
    <smtp_user><%= @smtp_user %></smtp_user>
    <smtp_password><%= @smtp_password %></smtp_password>
    <secure_smtp>yes</secure_smtp>
  </global>

  #install recipe
execute "run installer" do
  command "bash /opt/ossec/ossec-hids-master/install.sh"
  environment "SENDMAIL_CURL" => "yes"
  user "root"
  group "root"
end
```

### DIFF
--- a/src/os_maild/curlmail.c
+++ b/src/os_maild/curlmail.c
@@ -1,0 +1,123 @@
+/* Sendmail implementation using curl instead of writing it ourselves.
+ * Useful for sending SMTP traffic over authenticated hosts.
+ *
+ */
+
+#ifdef SENDMAIL_CURL
+#include <curl/curl.h>
+#include "maild.h"
+#include "mail_list.h"
+
+struct upload_status {
+  int lines_read;
+};
+
+static char *payload_text[7] = {NULL, NULL, NULL, NULL, NULL, NULL, NULL};
+
+static size_t payload_source(void *ptr, size_t size, size_t nmemb, void *userp) {
+  struct upload_status *upload_ctx = (struct upload_status *)userp;
+  const char *data;
+ 
+  if((size == 0) || (nmemb == 0) || ((size*nmemb) < 1)) {
+    return 0;
+  }
+ 
+  data = payload_text[upload_ctx->lines_read];
+ 
+  if(data) {
+    size_t len = strlen(data);
+    memcpy(ptr, data, len);
+    upload_ctx->lines_read++;
+ 
+    return len;
+  }
+ 
+  return 0;
+}
+
+int OS_Sendsms(MailConfig *mail, struct tm *p, MailMsg *msg) {
+  CURL *curl;
+  CURLcode res = CURLE_OK;
+  struct curl_slist *recipients = NULL;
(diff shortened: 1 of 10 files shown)


src/os_maild/curlmail.c
```diff
@@ -0,0 +1,123 @@
...
+
+  if(curl) {
+    char errbuf[CURL_ERROR_SIZE];
+    errbuf[0] = 0;
+    fprintf(stderr, "curl_easy_setopt() URL: %s\n", mail->smtpserver);
+    curl_easy_setopt(curl, CURLOPT_CAINFO, "/etc/ssl/certs/cacert.pem");
+    curl_easy_setopt(curl, CURLOPT_URL, mail->smtpserver);
+    curl_easy_setopt(curl, CURLOPT_ERRORBUFFER, errbuf);
+    curl_easy_setopt(curl, CURLOPT_IPRESOLVE, CURL_IPRESOLVE_V4);
+    curl_easy_setopt(curl, CURLOPT_DNS_SERVERS, "10.0.0.2,8.8.8.8,8.8.4.4");
```
nbuuck: Should these really be hard-coded rather than relying on the system-configured name servers?

sempervictus:
#1381 looks to be adding libsodium, if that PR is to be implemented, would it make sense to convert the mail functionality to a thinner internal implementation? Primarily asking because curl [does not come without concerns](https://github.com/ossec/ossec-hids), nor does any library, but the depth of opaque function calls relative to this code is probably a bit deeper than if the send/recv/fmt was implemented here, and crypto was the sole component being farmed out.

atomicturtle:
Superseded by #2195 (merged).

Thanks @alexbartlow — your curl-based TLS/AUTH SMTP approach landed on main
as USE_CURL=yes with auth_smtp/secure_smtp (see curlmail.c). Closing this
older PR in favor of that work.

PR #1381, opened 2018-03-15
Add Libsodium sodium support for sha256 filehashes

ddpbsd:
EXPERIMENTAL: Add support for sha256 checksums in syscheckd. It's optional for now under `USE_LIBSODIUM`.
If libsodium is enabled, sha1 is disabled. There's currently no way to choose it over sha256.

Only tested on OpenBSD and CentOS 7. Completely untested on Windows and other unix-like systems. Requires libsodium development packages on systems where developers are second class citizens.


src/config/syscheck-config.c
```diff
@@ -824,21 +866,27 @@ char *syscheck_opts2str(char *buf, int buflen, int opts) {
         CHECK_SIZE,
         CHECK_OWNER,
         CHECK_GROUP,
-	CHECK_MD5SUM,
+        CHECK_MD5SUM,
         CHECK_SHA1SUM,
         CHECK_REALTIME,
         CHECK_SEECHANGES,
+#ifdef LIBSODIUM_ENABLED
```
sempervictus: Having MD5 and SHA1 together makes some sense as both can have collisions, but dual-mode collisions are much less likely. SHA256 on the other hand is already computationally expensive, and not determined to have predictable collisions far as i know. Would it make more sense to
```
#ifdef LIBSODIUM_ENABLED
CHECK_SHA256SUM,
#else
CHECK_MD5SUM,
CHECK_SHA1SUM,
#end
```
so as to avoid the redundancy in processing and storage?

src/config/syscheck-config.c
```diff
@@ -824,21 +866,27 @@ char *syscheck_opts2str(char *buf, int buflen, int opts) {
...
         "perm",
         "size",
         "owner",
         "group",
-	"md5sum",
+        "md5sum",
         "sha1sum",
         "realtime",
         "report_changes",
+#ifdef LIBSODIUM_ENABLED
```
sempervictus: Same question as above

src/os_crypto/md5_sha1/md5_sha1_op.c
```diff
@@ -86,3 +91,99 @@ int OS_MD5_SHA1_File(const char *fname, const char *prefilter_cmd, os_md5 md5out
 
     return (0);
 }
+
+#ifdef LIBSODIUM_ENABLED
+int OS_Hash_File(const char *fname, const char *prefilter_cmd, struct hash_output *file_output, int mode)
```
sempervictus: Following along with the logic from above, SHA1 is supposed to be checked, but this function deals with MD5 and SHA256 only far as i see.

sempervictus (commented):
Please pardon my uninitiated status here, this entire thread of logic may stem from my ignorance of the platform's inner workings, but it looks like the internal representation of FIM construct is destabilizing in this PR with the conditional introduction of functions and structs based on a library dependency. Is it feasible to split this out into old/new implementations with exclusive behavior, or inclusively to have all hashes stored, but only the "hardest" one checked as available. The latter approach may be more forward-portable as it would permit named members of the hash_output struct to be added via preprocessor checks, and maintain a consistent calling convention, with the function meat also defined this way.

src/os_crypto/md5_sha1/md5_sha1_op.c
ddpbsd: It's been a long day, so I might be misunderstanding you. Bear with me and be gentle.
OS_MD5_SHA1_File was the original function. To make things easier for myself, OS_Hash_File was a copy of that (in the same file to, again, make things easier). Rather than deal with the directory naming issues and whatnot, I just wanted to get the functionality down.

src/config/syscheck-config.c
ddpbsd: I'm not against that, I was just continuing the tradition of using 2 hashes. I hope to replace md5 with blake2b (or whatever libsodium has as default at that point) in the future. md5 and sha256 give an analyst more opportunities to compare the hashes to "known bad" files out there. I've seen plenty of Indicators of Compromise reports still using md5.

src/os_crypto/md5_sha1/md5_sha1_op.c
sempervictus: Sorry for the confusion, it looks to me as though these changes would depend on both functions - since MD5 and SHA1 are always checked, which seems a bit redundant in the first place, but very much so if we run 3 (4 if MD5 happens once in each function being called) hash passes. If in fact the new function is being called instead, then it won't deal with the SHA1 piece, which still  needs to be verified.

src/config/syscheck-config.c
sempervictus: Good point, and some still use sha1, which sounds like "we might want them all." Maybe a semantic in which we store one hash and create other-algo-derived hashes on a change event for the alert would address that? Users could configure generation of event hashes in the fmt used by their malware engine, or just to pass the unk file to viper or whatever they're using for sample management to let another component deal with that rabbit hole.

src/os_crypto/md5_sha1/md5_sha1_op.c
ddpbsd: I thought I had made it so only OS_Hash_File() gets called if libsodium support is added. I'll have to double check.

ddpbsd:
My original PR message wasn't very detailed, so I'll add some more details here.
My roadmap for this is basically:
3.0: continue with md5/sha1 - It's too close to release to merge this IMO.
3.1: default to md5/sha1, allow users to test this out if they want.
3.2: move to md5/sha256 (or blake2b/sha256 if I get saucy). Remove support for sha1 entirely.

I don't personally want to keep sha1 around that long, but I also don't want to cause too many issues for users. This seemed like a decent compromise. (Or if sha1 is necessary for some reason, I can roll it into the newer methods, but I'm hoping not to.)

I wanted to minimize changes in analysisd, the piece of the puzzle that compares the new hashes to the old. I have an idea on how to make this whole transition easier in the future, but I'm not sure it's good enough and it conflicts with my idea of minimizing analysisd changes.

For the transition, right now there will be a lot of alerts (sha1 hash vs sha256 hash). I could add a check into analysisd to compare the length of the hashes. If the old hash is shorter, don't alert (there will still be an alert if the md5 hash doesn't match). I'm not entirely sure I like this idea, but it's the best one I've come up with yet.

Hopefully this explains the choices I've made a bit better.

ddpbsd:
I listened to a podcast that mentioned bro ids only does sha1 and md5 for its network file hashing. So I'm thinking we'll have to leave one or both of these algorithms for things like that. When I'm not typing into a tiny keyboard, I'll revise my road map a bit.

sempervictus:
Far as interoperability goes, sha1 probably has to stay for some time to come. However, md5 collisions aren't that hard to create, and I for one might abuse this on a pentest if I know they're relying on md5 for a world writable privileged context binary/lib. I'll hunt you down at the conference, maybe we can, umm, hash out details.

sempervictus:
If we are adding Virgil's crypto, do we need libsodium, or could we leverage their imports for the functionality to reduce weight?

ddpbsd:
During the initial phone call with Virgil I mentioned libsodium. IIRC they thought it was a good idea.

ddpbsd:
`src/analysisd/format/to_json.c` has a bunch of goodies that need to be updated.

atomicturtle:
Ended up doing this without libsodium in 4.x

PR #1400, opened 2018-04-16
Fix SID syntax for group membership changes

jjrbg:
SIDs in logs from Active Directory are not contained within curly brackets, nor are they prefixed with a %. Perhaps this is from old NT days. This is an example log:

2018 Mar 01 00:00:00 WinEvtLog: Security: AUDIT_SUCCESS(4728): Microsoft-Windows-Security-Auditing: (no user): no domain: domaincontroller.lan.local: A member was added to a security-enabled global group. Subject:  Security ID:  S-1-5-21-0000000000-0000000000-000000000-00012  Account Name:  admin_account  Account Domain:  LAN  Logon ID:  0x11cc0174  Member:  Security ID:  S-1-5-21-0000000000-000000000-000000000-00675  Account Name:  CN=added_account,OU=lan,DC=lan,DC=local  Group:  Security ID:  S-1-5-21-0000000000-0000000000-000000000-512  Group Name:  Domain Admins  Group Domain:  LAN  Additional Information:  Privileges:  -

### DIFF
--- a/etc/rules/msauth_rules.xml
+++ b/etc/rules/msauth_rules.xml
@@ -499,7 +499,7 @@
 
   <rule id="18218" level="5">
     <if_sid>18207,18208</if_sid>
-    <regex> ID:\s+%{S-1-1-0}</regex>
+    <regex> ID:\s+S-1-1-0</regex>
     <description>Everyone Group Changed</description>
     <group>group_changed,win_group_changed,</group>
     <info>http://support.microsoft.com/kb/243330</info>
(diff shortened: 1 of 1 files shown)


ddpbsd:
Would it be unreasonable to cover both? Without knowing when this changed(?), I hesitate to remove backwards compatibility.
Something like:
```
<regex> ID:\s+%{S-1-1-0}| ID:\s+S-1-1-0</regex>
```

jjrbg:
Fair comment. I've resubmitted.

Issue #1402, opened 2018-04-18
Real-time file monitoring stops working if several files are encrypted at the same time


billy-shears:
1. Deploy HIDS agent on Windows 2008 server

2. Disable auto_ignore option in ossec server configuration and restart ossec
/etc/init.d/ossec restart

3. Add real-time monitoring for C:\Data directory in agent ossec.conf
<directories check_all="yes" realtime="yes">C:\Data</directories>

4. Create C:\Data\Test folder and 200 files inside Test and restart ossec agent service on Windows 2008 server

5. Add real-time monitoring for C:\Data directory in agent ossec.conf and restart agent service
<directories check_all="yes" realtime="yes">C:\Data</directories>

6. Check that C:\Data directory is being monitored in real time by ossec agent and wait until real-time file monitoring starts
agent ossec.log

2018/04/13 15:08:00 ossec-syscheckd: INFO: Monitoring directory: 'C:\Data', with options perm | size | owner | group | md5sum | sha1sum | realtime.
..
2018/04/13 14:01:24 ossec-syscheckd: INFO: Real time file monitoring started.

7. Encrypt C:\Data\Test folder and files
  a. Select C:\Data\Test folder
  b. Right-click the file or folder and select Properties.
  c. On the General tab, click the Advanced button.
  d. Check the box for the "Encrypt contents to secure data" option.
  e. Click OK and then Apply.
  f. Select 'Apply changes to this folder, subfolders and files' and click on OK

8. Check the error in ossec.log and if HIDS events are generated when the directory and files are encrypted

Actual Behavior: Real-time file monitoring stops working if several files are encrypted at the same time
Expected Behavior: Real-time file monitoring should not stop working if several files are encrypted at the same time

billy-shears:
This error appears in agent ossec.log file and after that real-time file monitoring stops working:
`2018/04/13 14:15:30 ossec-syscheckd: ERROR: real time call back called, but 0 bytes.`

billy-shears:
Additional info about this defect:

It depends on the number of files encrypted at the same time. Following the steps to reproduce above, if only first 25 files from Test folder are encrypted at the same time the defect doesn't happen

```
Lab-HIDS:~# tail -f /var/ossec/queue/syscheck/\(W2008\)\ 10.5.1.2-\>syscheck
!++49:33206:0:0:xxx:xxx !1523633883 C:\Data/Test/File1.txt
[...]

PR #1207, opened 2017-07-24
Avoid EOL conversion of received files in the Windows receiver

QuantumHeart:
See issue #1205.

Issue #1205, opened 2017-07-20
Pushing merged.mg to Windows agents fails due to EOL conversion


QuantumHeart:
Most likely related to:
https://groups.google.com/forum/#!topic/ossec-list/OsUrHR4ykuo

This bug has been observed in OSSEC 2.8.3 and 2.9.0. 2.9.1 is not available for Windows yet, but looking at the changelog it is most likely this bug still exists.

Situation:
An OSSEC server combines all files in `/var/ossec/etc/shared` into a single file known as `merged.mg`. Since the server uses Linux, all lines in `merged.mg` end with a line feed (LF, hex code 0A).

`merged.mg` is pushed to agents when they connect. A **Windows** agent receives and reconstructs the file (send over UDP 1514) and an MD5 hash, and verifies the MD5 hash of the file to see whether the transfer was successful.

Expected outcome:
Assuming the transfer went without problems, the Windows agent reconstructs the file as it was on the server to allow proper verification using the MD5 hash that was also received. Therefore, line feed will be used as an end-of-line character. The file is reconstructed exactly as it was on the server, and the MD5 sum verifies.

Actual outcome:
Assuming the transfer went without problems, the Windows agent reconstructs the file using carriage return (CR, hex code 0D) **and** a line feed (LF, hex code 0A) as newline characters. The result is that the MD5 verification will always fail, since the MD5 hash generated by the Linux server (using LF newlines) will never match that of the MD5 generated by the Windows agent (using CRLF newlines).

Suggestion:
Change the verification process of the agent.

1. Receive the data and the MD5 sum.
1. Reconstruct the file as-is. Do not do any conversion. If this is not possible (due to the way the server sends files), reconstruct the file using all possible EOL combinations and verify all variations in the next step.
1. Verify the MD5 hash.
1. When the MD5 hash matches, do an EOL conversion to the target platform of the agent. For Windows, this would be a conversion to CRLF.

nbuuck:
Confirmed that MD5 is failing consistently for Windows agents. Set `windows.debug` to 1 in `local_internal_options.conf` of a Windows agent, restarted the agent, and observed:

> 2017/07/22 19:29:59 ossec-agentd: Failed md5 for: shared/merged.mg -- deleting.

nbuuck:
I tried to verify this by reviewing the code but between the project structure (receiver.c versus receiver-win.c), build automation, and minimal code comments, I'll likely have to review in a debugger to verify.

[...]

PR #1413, opened 2018-05-03
V2.9.4 - More backporting

ddpbsd:
Issue #1402 and companion pull request #1409 
Pull request #1207 and companion issue #1205


ddpbsd:
Added the `ossec-slack.sh` fix.

Issue #1421, opened 2018-05-16
ossec-slack.sh defunct


amh5366:
Hey!

We encountered a problem with our slack integration that I thought it'd be best to bring to light. The script ossec-slack.sh tries to read and access files in /var/ossec/active-response/log/active-responses.log and /var/ossec/active-response/log/alerts/alerts.log. These files do not exist here, and the script fails as a result.

The files are located in /var/ossec/logs/* respectively. I should note that we install ossec via rpms!

PR #1422, opened 2018-05-17
Fix the log location in the ossec-slack AR script.

ddpbsd:
The script looks for `${PWD}/../logs`, but that does not exist.
It should look for `${PWD}/../logs` instead since it is being runfrom ossec/active-response/bin.

Reported by amh5366 in issue #1421


foygl:
I had to revert this change to get the integration to work in 3.0.0

PR #1430, opened 2018-05-29
Update local_rules.xml

tsia:
1.1.1.1 actually is used. according to RFC 5737 "The blocks 192.0.2.0/24 (TEST-NET-1), 198.51.100.0/24 (TEST-NET-2), and 203.0.113.0/24 (TEST-NET-3) are provided for use in documentation."


ddpbsd:
Good catch, thanks!

PR #1435, opened 2018-06-10
Added script for blocking/unblocking IP on windows firewall

tommed:
Been using this on my Windows servers for some time now, but thought it would be nice to contribute it back to the community. This uses `netsh` to add the temporary blocks into the Windows Firewall for potentially malicious callers, then removes them again after the configured interval.


tommed:
Any other points I've missed?

Issue #1448, opened 2018-07-03
Commits merged to master, missing from OSSEC 2.9.4 for JSON output


reyjrar:
There are a few commits that enhance the JSON output dramatically that I have to patch-in every release which should be included in the official releases:

* 4985d87f39aafbd816e5bfbf93ee6050332c77f6
* 7e094aeeeb1e0f55dca335cdae8237f6d50b99e3
* 278a66ea0d4334970a552d9d29f63bcdea9868f4

These commits output the entire alert event as JSON.  There's a lot of valuable data we're not passing to the alerts.json and ZeroMQ output without these commits.

PR #1449, opened 2018-07-03
Add more fields to JSON output.

reyjrar:
We already have parsed these elements and used them in the rules, so why
not pass them along to JSON consumers.  This gives users more
flexibility to analyze their alerts in a system like ElasticSearch
without having to reparse or remap this data.  This will hopefully
deduplicate slower Logstash type configs.


reyjrar:
This compresses the referenced commits #1448 into a single patch against 2.9.4.  Since there's no 2.9.5 branch yet.

PR #1453, opened 2018-07-11
set nullsessionpipes to check emptiness

Bob-Andrews:
Recommended value for member server is <blank>.


atomicturtle:
Thanks again for your hard work on these!

PR #1464, opened 2018-07-19
BUGFIX: remove default value for authpass

phamvuong:
authpass is originally set as NULL
If user does not specify option "-P" but then incsecure mode is used.
However the authpass is set to some default value by mistake which
forces user to always use password

### DIFF
--- a/src/os_auth/main-client.c
+++ b/src/os_auth/main-client.c
@@ -246,11 +246,6 @@ int main(int argc, char **argv)
         exit(1);
     }
 
-
-    if(authpass == NULL) {
-        authpass = AUTHDPASS_PATH;
-    }
-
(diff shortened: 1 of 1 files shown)


jsilverman-blispay:
When will this patch become available in precompiled binaries? (e.g. here: https://updates.atomicorp.com/channels/ossec/ubuntu/pool/main/o/ossec-hids-agent/)

Or is this question purely for the Atomicorp repo maintainers?

Thanks

atomicturtle:
It'd probably end up in atomic-testing sometime this week as a snapshot build. There is no automation for that yet. Barring that, when we nail down any last commits for 3.0.1

phamvuong:
@jsilverman-blispay Ossec 3.1 has been released including this patch

PR #1505, opened 2018-08-17
Remove the automated reports options

ddpbsd:
The automated reports run by `ossec-monitord` adds more complexity than I'd like, for very little gain.
Pretty much every OSSEC server will have cron(8), so running a report on a schedule shouldn't be difficult.


atomicturtle:
Automated monitord reports are still a supported feature on main and have
seen recent maintenance (SMTP auth, worker/threading hardening). Prefer
keeping the built-in path rather than removing it in favor of cron-only.

PR #1531, opened 2018-09-19
Added Rootchecks

Bob-Andrews:
Rootchecks for Debian 7+8


Bob-Andrews:
This check produces one Alert (./rootcheck -i...) per rc*.d folder containing the file (e.g. /etc/rc1.d/S01avahi-daemon and /etc/rc2.d/S01avahi-daemon). I am not sure if this is a problem or not. 

And it is searching for exactly runlevels and priorities given from the cis-benchmark (e.g. f:!/etc/rc2.d/S01iptables-persistent;). Is this too accurate?

PR #1536, opened 2018-09-21
Add -f foreground flag support

atomicturtle:
New feature for authd, the previous behavior was to run the daemon in
the foreground. This changes the default to run in the background, and
implements the -f flag used on the other daemons to force foreground
mode.


ddpbsd:
I think we should not commit the zlib changes, but otherwise this looks good to me.

PR #1543, opened 2018-10-01
Added new rules

Bob-Andrews:
New Rules for USB detection and Windows firewall

### DIFF
--- a/etc/rules/linux_usbdetect_rules.xml
+++ b/etc/rules/linux_usbdetect_rules.xml
@@ -1,0 +1,43 @@
+<!-- OSSEC USB-detection Rule for Linux - https://www.thomas-krenn.com/de/wiki/Ubuntu_Syslog -->
+
+<group name="linux, usb,">
+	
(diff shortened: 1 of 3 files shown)


ddpbsd:
The rule IDs are in the user defined range (100000+).
Please use the following ranges, and update the `/doc/rule_ids.txt` file:

`linux_usbdetect_rules` 53600 - 53625

`ms1016_usbdetect_rules`: 53626 - 53630

`ms_firewall_rules.xml`: 53631 - xxxxxx (end of the range to be determined)

Also, this may be nitpicky, but the other rules are using 2 spaces instead of tabs (or 8 spaces?).

Bob-Andrews:
Corrected Tabs to Spaces and changed IDs. Update /doc/rule_ids.txt is in request #1546 . ms_firewall_rules need only a range of 53631 to 53654. I changed it in rule_ids.txt to a range of 53631 - 53699 because it seems to me, that other ranges end at xxx50 or xxx99,

PR #1547, opened 2018-10-05
Corrected IDs to a non user defined range

Bob-Andrews:
Corrected IDs from psad_rules, sysmon_rules and unbound_rules.xml from ids in user defined range to ids below 100000. Added the new ranges to doc/rule_ids.txt.


ddpbsd:
Wow, good catch. I'll correct the tests in another PR.

PR #1574, opened 2018-11-16
Added rules/decoder for mhn

Bob-Andrews:
Added Range, Rules and Decoder for Cowrie and Dionaea in Modern Honeypot Network. 
Did not add the mhn-json.log to the ossec.conf. Added a hint to the rule files and decoder instead. 
Not sure if this rules are to special for including it directly.


aquerubin:
The regexp are IPv4 centric.  Can you make them IPv6 friendly?

Tony

PR #1583, opened 2018-11-24
Remove double hyphen in comment

### LINKED ISSUE #1582: ossec-hids/etc/ossec-agent.conf contains double hyphen in comment
This is illegal as per https://www.w3.org/TR/REC-xml/#sec-comments

> [...]the string " -- " (double-hyphen) must not occur within comments.

The occurrence of a double hyphen may prevent proper processing of the file by various XML libraries, not the least of them Python's lxml module, which is used by Ansible.

mwmahlberg:
Fixes #1582

### DIFF
--- a/etc/ossec-agent.conf
+++ b/etc/ossec-agent.conf
@@ -6,9 +6,9 @@
   </client>
 
   <syscheck>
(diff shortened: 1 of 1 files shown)


ddpbsd:
Thanks!

Issue #1509, opened 2018-08-27
IPExist function available but never used leading to duplicated IPs with authd


tanderson3:
In PR#1161 - we can see that the function IPExist has been set back:
https://github.com/ossec/ossec-hids/pull/1161

But this function is never used and it's still possible to duplicate IPs using the authd mechanism.
It had been enabled in that patch:
https://github.com/ossec/ossec-hids/pull/1162/files#diff-c1d16a58d2ed41485e829b7ff16b2ecfR495
Then removed with this patch:
https://github.com/ossec/ossec-hids/commit/352e740aa7097f0e0be7a43f938a96d225a2f552

Would it be possible to re-enable that feature since the function seems to come back in code?

tanderson3:
Any plan to get that "IP duplicate check" function back in a next release?

tanderson3:
Sounds like it's working with that fix now. Thanks!
Do you think this could be merged anytime soon?

ddpbsd:
I think @atomicturtle is planning on a new release in January, but don't quote me on that.

tanderson3:
Ok, nice. Thanks Dan!

PR #1603, opened 2018-12-19
Use IPExist to check for duplicate IP addresses.

ddpbsd:
Lightly tested, seemed to work for me.
Issue #1509


atomicturtle:
Will follow up with the NAT testing on this on my side during packaging.

PR #1607, opened 2018-12-19
Expanding JSON Compliance

atomicturtle:
Expanding JSON output for the compliance functions. Prep work for
root/syscheck_control.

Original by Pedro S


snaow:
Thanks!

PR #1614, opened 2018-12-25
First attempt at a 64bit Windows binary.

ddpbsd:
Remove lua, it was causing issues during the build.

I can't get the ui to run on the 1 test machine I have, but it could just be throwing a tantrum right now.
The installer still seems to identify as a 32bit app.
I haven't looked at hooking it into travis-ci yet.
I'm not entirely sure how to test whether everything else works. So I guess it's time for users to take a peek?

Addresses issue #301


ddpbsd:
Need to figure out the 64bit registry stuff: https://docs.microsoft.com/en-us/windows/desktop/WinProg64/shared-registry-keys
More info: https://stackoverflow.com/questions/10533421/accessing-64-bit-registry-from-32-bit-application

PR #1641, opened 2019-01-16
Improved dotests.sh output 

jubois:
- Changed output to diff --side-by-side style
- Runs all tests and set exit code to 0 only if all tests passed (else 1)

### DIFF
--- a/contrib/logtesting/dotests.sh
+++ b/contrib/logtesting/dotests.sh
@@ -1,5 +1,22 @@
 #!/bin/sh
 
+hostname=`hostname`
+hostname melancia
+
+cleanup() {
+    hostname $hostname
+    rm -f ./tmpres
+}
+
+trap "cleanup" INT TERM EXIT
+exitcode=0
+
+if diff --help 2>&1 | grep -q -- --color; then
+    diff_cmd='diff --color'
+else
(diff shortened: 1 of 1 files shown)


contrib/logtesting/dotests.sh
```diff
@@ -19,17 +25,22 @@ for i in ./*/log; do
...
+        echo
+        echo
         echo "**ERROR: Unit testing failed. Output for the test $i failed."
-        echo "== OLD OUTPUT: =="
-        cat $idir/res
-	echo "== NEW OUTPUT: =="
-        cat tmpres
-        echo "** ERROR: Exiting."
+        echo "== DIFF OUTPUT (reference | new): =="
+        diff -Nay $idir/res tmpres
```
ddpbsd: My `diff` doesn't have a `-y` flag.

jubois: I checked with the virtual ossec appliance and the ``-y `` exists...
What is your distribution ?

ddpbsd: OpenBSD in this case. I haven't checked others.

ddpbsd:
It doesn't work on OpenBSD anyway. I'll have to peek at that later. Thanks for the updates!

PR #1646, opened 2019-01-18
Added dangerous commands/background activity catching rules

Bob-Andrews:
Added rule for dangerous powershell commands +
Added rules with level 2 for catching less critical background activity commands like powershell starting or error printing for less wrong mail warnings.


atomicturtle:
Excellent work on this one!

PR #1652, opened 2019-01-22
Added feature pcre2

jubois:
- Replaces both OSRegex and OSMatch engines with pcre2 engines.
        - They both converted to pcre2 patterns
        - No more bugs (had to do ``p = OSRegex_Execute(...); p++``
        to get correct return value).
    - Added new XML tags to directly use pcre2 patterns :
        - pcre2 (to replace regex)
        - match_pcre2
        - program_name_pcre2
        - prematch_pcre2
        - srcgeoip_pcre2
        - dstgeoip_pcre2
        - srcport_pcre2
        - dstport_pcre2
        - user_pcre2
        - url_pcre2
        - id_pcre2
        - status_pcre2
        - hostname_pcre2
        - extra_data_pcre2
    - Added util program to convert OSMatch/OSRegex patterns to pcre2
    ones.
    - Added script contrib/ossec-pcre2-config.pl to update rules &
    decoders
    - Had to change ``Config->hostname_white_list`` type to ``char **``
    because OSMatch no longer have the ``patterns`` field.
        - Update install.sh script due to the change


jubois:
The compilation of libpcre2 fails with clang, (pcre2-10.32/libtool: line 10756: clang: command not found) at link time.
Cannot figure out why. Any help will be much appreciated !

ddpbsd:
Hi @jubois 
Thanks for submitting this PR, it looks like a lot of work!
I'm pretty busy this week, so it might take me a while to really get a good look at it. But we are definitely interested!

jubois:
FYI this release corrects several Ossec issues:

**These regex should not match:**

./ossec-regex '^\D\D\.*\d\d'
!!sk8er
+OSRegex_Execute: !!sk8er
+OS_Regex       : !!sk8er

./ossec-regex '\d\d\.*\D\D'
111111111a111111111
+OSRegex_Execute: 1111111111a111111111
+OS_Regex       : 1111111111a111111111

./ossec-regex '\D\D\.*\D\d'
aaaa
+OSRegex_Execute: aaaa
+OS_Regex       : aaaa

**Capture groups error**

Modified ossec-regex to print captured group

./ossec-regex '(\S+).$'
jones_b.
+OSRegex_Execute: jones_b.
+OSRegex_Execute captured: jones_b.

Result should be jones_b (without « . »)

Should I do "issues" for these bugs?

Defaults Ossec modifiers behavior is strange (not working after simple chars), and so is greediness (modifiers are only greedy when there is nothing after). This PR do the exact same thing with ossec formated rules.

jubois:
Thank you for your compliments ;-)

atomicturtle:
I thought I'd relay this comment from @jrossi in our slack channel (let me know if you'd like an invite! scott@ossec.net):

"Pcre2 is the best thing I have seen in weeks!!  Awesome sauce!  Not that I have any dog in this fight, but with all the amazing transition work i would invert the xml so that regex is pcre2 by default and the Perl translation script  takes regex and transforms it from regex_legacy.  Still awesome contrib for the project!"

Issue #1664, opened 2019-02-08
Syslog ISO timestamp events with optional fraction of second are skipped by pre-decoder


MangyCoyote:
Related to   #1032 

Syslog events like:
2019-02-06T16:15:28.1+00:00 ...
2019-02-06T16:15:28.22+00:00 ...
2019-02-06T16:15:28.333+00:00 ...
etc. are not processed by [pre]decoder.

Supported formats at the moment are only:
2019-02-06T16:15:28+00:00 ... simple iso
2019-02-06T16:15:28.666666+00:00 ... due to rsyslog.

Expand fraction of a second to be optional up to reasonable time granularity (6 digits).

ddpbsd:
Thanks for the issue and pull request!

MangyCoyote:
You are welcome. I have to express my thank you for this tool  on the other hand 👍

atomicturtle:
Wow this solves a bunch of issues I was having with pre-decoding some medical device logs. Thank you so much for this!

MangyCoyote:
Extremely glad the change helped you in such critical area where medical devices are!

PR #1665, opened 2019-02-08
#1664 - Syslog ISO timestamp events with optional fraction of second …

MangyCoyote:
…are skipped by pre-decoder


MangyCoyote:
This patches allows 6 optional fraction of a second. Separator is also optional: either dot or comma

PR #1685, opened 2019-03-31
Copy /etc/resolv.conf to /var/ossec/etc.

aquerubin:
Copy the resolver config to the chroot environment to fix getaddrinfo errors.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -23,6 +23,7 @@ OSSEC_USER_REM?=ossecr
 
 INSTALL_CMD?=install -m $(1) -o $(2) -g $(3)
 INSTALL_LOCALTIME?=yes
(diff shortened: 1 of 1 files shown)


ddpbsd:
Should we copy the resolv.conf in ossec-control when starting ossec, in case things change?

aquerubin:
While it may be a good idea for other reasons, it's not absolutely necessary to fix the getaddrinfo errors that occur on initial installation and setup on certain OS.  Although the resolv.conf files may get out-of-sync, it does not automatically imply breakage will result.  And doing so means an ossec config file changes on every startup.

Perhaps we should do this only on agents which are possibly more mobile than managers?

ddpbsd:
That makes sense.

PR #1697, opened 2019-04-21
Updated IPv4-dependent regexp in ownCloud decoders.

aquerubin:
Updated IPv4-dependent regexp in ownCloud decoders.

### DIFF
--- a/etc/decoder.xml
+++ b/etc/decoder.xml
@@ -3123,21 +3123,21 @@ s=2 SFwdQ=0 SDupQ=0 SErr=0 RQ=2 RIQ=0 RFwdQ=0 RDupQ=0 RTCP=0 SFwdR=0 SFail=0 SFE
 <decoder name="owncloud-failed1">
   <parent>owncloud</parent>
   <prematch>Login failed: user </prematch>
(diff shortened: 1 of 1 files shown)


iasdeoupxe:
@aquerubin @ddpbsd Are you sure that this is a correct change for this? If you check the examples in the decoder below you can see that the ``\S+`` operator which is "anything but space" will probably match more then it should:

https://github.com/ossec/ossec-hids/blob/b7cfd4db56b15632a33c204f9dd494a43e82f664/etc/decoder.xml#L3093-L3104

ddpbsd:
@iasdeoupxe Is there a specific instance of `\S` you're referring to?

iasdeoupxe:
I basically mean the following changed in this PR:

```
<regex offset="after_prematch">^'(\w+)' , wrong password, IP:(\S+)</regex>
```

in the log snippet posted above i can see e.g.:

```
wrong password, IP:127.0.0.1","level"
```

IIUC the ``\S+`` means "everything but space" where not only the IP but also the following text will be extracted into the IP or am i wrong?

aquerubin:
Yikes!  You’re right.  Will fix.

Tony

iasdeoupxe:
Initial fix in https://github.com/ossec/ossec-hids/pull/1724, did a minor update in https://github.com/ossec/ossec-hids/pull/1725

PR #1705, opened 2019-05-06
fix bug

wcc526:
fix bug,append will duplicate the /etc/init.d/ossec file

### DIFF
--- a/contrib/specs/agent/ossec-hids-agent.spec
+++ b/contrib/specs/agent/ossec-hids-agent.spec
@@ -69,7 +69,7 @@ done
 ################################################################################
 # Create OSSEC /etc/init.d/ossec file
 #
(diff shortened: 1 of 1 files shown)


ddpbsd:
The `>>` appends to the file. A single `>` over writes the file.
Is that what we want? (I don't know anything about the RPM stuff, so maybe this is what is desired?)

wcc526:
https://stackoverflow.com/questions/48505456/creating-init-script-from-shell-script

ddpbsd:
I understand what the shell code itself does, but I don't know if that's what should happen in the RPM side of things.
Plus, the spec file is for ossec 1.3. I doubt it's relevant to anything anymore.

wcc526:
>> append file, it will cause the duplicate content of /etc/init.d/ossec if the ossec reinstall

PR #1709, opened 2019-05-08
Added Cloudflare active-response script.

almirb:
It adds (and removes) an IP to Cloudflare IP block list.

### DIFF
--- a/active-response/cloudflare-ban.sh
+++ b/active-response/cloudflare-ban.sh
@@ -16,7 +16,7 @@
 #  <active-response>
 #     <command>cloudflare-ban</command>
 #     <location>server</location>
-#     <rules_id>31151,31163</rules_id>
+#     <rules_id>31151,31152,31153,31154,31163</rules_id>
 #     <timeout>43200</timeout>
 #  </active-response>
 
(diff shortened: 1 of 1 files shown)


featzor:
> It adds (and removes) an IP to Cloudflare IP block list.

Hello @almirb do you know from top of your head the permissions required at Cloudflare for the integration?

featzor:
@almirb @ddpbsd hello again got this to work but the timeout don't seem to work the list just keep grow and grow. Here's my settings in ossec.conf.

`  <active-response>
     <command>cloudflare-ban</command>
     <location>server,000,001</location>
     <level>5</level>
     <timeout>43200</timeout>
  </active-response>`

featzor:
@almirb @ddpbsd The `/cloudflare-ban.sh delete` do not seem to work. The count keep grow and checking the active-response log for example the last delete entry:

`Sat Jan 25 22:44:56 CET 2020 /var/ossec/active-response/bin/cloudflare-ban.sh delete - 84.107.86.107 1579945449.41832 31101`

If I search for the IP above in the cloudflare IP Access Rules tools its still there.

featzor:
@almirb @ddpbsd found the issue command **jq** used to retrieve the id from the result array is not posix hence why the command do not work.

PR #1712, opened 2019-05-16
PCRE2 decoders conversion

jubois:
All decoders have been converted from legacy format to PCRE2


ddpbsd:
I don't have time to go over all of these right now, so I'm going to merge them. We can fix any fallout in-tree.
Thanks for all of this!

jubois:
You're welcome :-)

PR #1725, opened 2019-06-01
Additional ownCloud decoder fix

iasdeoupxe:
Follow-up of: https://github.com/ossec/ossec-hids/pull/1697 + https://github.com/ossec/ossec-hids/pull/1724

The decoder.xml shows the following two different lines, seems some ownCloud versions had a different (broken) syntax:

```
(Remote IP: '127.0.0.1')",
```

and

```
(Remote IP: '127.0.0.1)",
```

### DIFF
--- a/etc/decoder.xml
+++ b/etc/decoder.xml
@@ -3130,7 +3130,7 @@ s=2 SFwdQ=0 SDupQ=0 SErr=0 RQ=2 RIQ=0 RFwdQ=0 RDupQ=0 RTCP=0 SFwdR=0 SFail=0 SFE
 <decoder name="owncloud-failed2">
   <parent>owncloud</parent>
   <prematch>Login failed: </prematch>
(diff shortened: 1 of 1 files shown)


iasdeoupxe:
Based on the discussion in https://github.com/wazuh/wazuh-ruleset/pull/381 this probably need to be the following instead:

```
<regex offset="after_prematch">^'(\w+)' \(Remote IP: '(\S+)'|^'(\w+)' \(Remote IP: '(\S+)\)</regex>
```

aquerubin:
How widespread is this broken syntax logging problem in Owncloud and is it fixed in the latest versions?

Tony

iasdeoupxe:
TBH i don't know how widespread this broken logging is but can confirm that its fixed in all 10.2 which i'm currently running.

aquerubin:
Actually, this patch has no real effect.

Tony

iasdeoupxe:
https://github.com/ossec/ossec-hids/pull/1726 should be the final version now.

PR #1733, opened 2019-06-18
Check return values for unlink(2) calls

ddpbsd:
This isn't all of the unlink(2) calls, but just a few. Hopefully more to come later.

I'm running it now, but it hasn't run for very long. If I don't make the PR now though, I'll forget.


ddpbsd:
No errors after running over night on OpenBSD amd64 and Centos 7 amd64.

PR #1738, opened 2019-07-05
Set PCRE2_SYSTEM to no by default.

ddpbsd:
pcre2 has been a problem for people. So instead of relying on them
to download the tarball and extract it in the correct place, let's
default to using a system provided pcre2 library.
Disable JIT by default on OpenBSD. It would be a bunch of work to
make it actually work, so skip it for now.

Documentation updates should happen.


ddpbsd:
`PCRE2_SYSTEM` has been set to no for travis due to the old version of Ubuntu.

atomicturtle:
Note for later docs update, JIT is going to flag on a lot of memory protection systems (grsecurity, etc). just the nature of JIT.. it is an exploit vector

PR #1751, opened 2019-07-23
Fixed missing same_source_ip in rule 11306

binrush:
I do not see any reason not to have `<same_source_ip />` in bruteforce rule. Please correct me if I wrong


atomicturtle:
Excellent catch! you're absolutely right. Just about any time you're using a frequency rule looking at a key (IP, filename, user, etc) it should be pinned to the same key.

PR #1752, opened 2019-07-24
Verify Googlebot

bchavet:
In order to verify a googlebot crawler, Google recommends doing a reverse DNS lookup, and then a forward DNS lookup to verify that

1. The host is in the googlebot.com domain
2. The forward and reverse lookups match.

https://support.google.com/webmasters/answer/80553

### DIFF
--- a/src/analysisd/compiled_rules/generic_samples.c
+++ b/src/analysisd/compiled_rules/generic_samples.c
@@ -11,6 +11,9 @@
 #include "shared.h"
 #include "config.h"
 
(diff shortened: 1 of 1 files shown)


bchavet:
This is a patch that I wrote years ago, and have been using ever since. I thought I had shared it but couldn't find it anywhere, so here it is now.

atomicturtle:
This is fantastic, I actually implemented the exact same thing in mod_security :P Thanks for the contribution here!

PR #1780, opened 2019-10-16
Implement multi-line collection for indented logs

sempervictus:
Multiple logging implementations utilize tabs or spaces to prefix
lines following the top, inindented line starting the entry. The
existing multi-line approach does not deal well with variable log
lengths using these indented sub-entries as it extracts lines by
count, not content.

This commit implements a modified postgresql log reader as a
generic multi-line parser sending the current buffer downstream
when it starts a new log entry (line does not start with ' ' or
'\t'), or encounters an empty line.

### DIFF
--- a/src/logcollector/read_multiline_indented.c
+++ b/src/logcollector/read_multiline_indented.c
@@ -1,0 +1,120 @@
+/* Copyright (C) 2019, Semper Victus LLC
+ * Copyright (C) 2009 Trend Micro Inc.
+ * All rights reserved.
+ *
+ * This program is free software; you can redistribute it
+ * and/or modify it under the terms of the GNU General Public
+ * License (version 2) as published by the FSF - Free Software
+ * Foundation.
+ */
+
+/* Read indentend multi line logs */
+
+#include "shared.h"
+#include "logcollector.h"
+
+
+/* Read multi line indented log files */
+void *read_multiline_indented(int pos, int *rc, int drop_it) {
+    size_t str_len = 0;
+    char *p;
+    char str[OS_MAXSTR + 1];
+    char buffer[OS_MAXSTR + 1];
+    /* Zero buffer and str */
+    buffer[0] = '\0';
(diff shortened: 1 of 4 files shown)


sempervictus:
ping @atomicturtle @ddpbsd  - could you guys please take a peek @ this? Also thinking for the actual config, we could use `multi-line:0` to keep existing format... thoughts/comments/fixes/profanity all welcome.

src/logcollector/read_multiline_indented.c
```diff
@@ -0,0 +1,120 @@
...
+#ifdef WIN32
+        if ((p = strrchr(str, '\r')) != NULL) {
+            *p = '\0';
+        }
+#endif
+        /* Look for empty string */
+        if ((str_len <= 1) || (str[0] == '\r')) {
+            /* Send existing data if any in buffer */
+            if (buffer[0] != '\0') {
+              if (SendMSG(logr_queue, buffer, logff[pos].file, LOCALFILE_MQ) < 0) {
```
ddpbsd: The other `read_*` files check to make sure `drop_it == 0` before sending to the queue.
I only see instances of `drop_it == 1` for Windows though, so it may not be necessary.
But it would quiet the following build warning:
```
logcollector/read_multiline_indented.c:18:53: warning: unused parameter 'drop_it' [-Wunused-parameter]
void *read_multiline_indented(int pos, int *rc, int drop_it) {
                                                    ^
1 warning generated.
```

sempervictus:
@ddpbsd: does that make the compiler happy?

ddpbsd:
Sorry for the delay. That warning is gone. Thanks for pull request!

sempervictus:
Thanks Dan, lets see if it eats cats now.
Any thoughts on syntactically wiring as multi-line:0?

ddpbsd:
I don't personally have a preference. I think the current implementation is more intuitive.

PR #1787, opened 2019-10-30
Adding lighttpd decoder/rules

Nono-m0le:
This is my first attempt, sorry if it's not the way it is supposed to be done.
- I didn't had any licences/information on the rules as I personnaly use it on the local_rules.xml
Obviously, this can freely reuse in any condition (WTFPL Licences)
Feel free to correct if needed or merge with another set of rules ?
- Rule id / level can be freely edited as needed (id="110000" / level="10")

Happy to get any feedback on this.

### DIFF
--- a/etc/decoder.xml
+++ b/etc/decoder.xml
@@ -1954,7 +1954,7 @@ Jan  8 19:32:41 tp.lan dropbear[15165]: Pubkey auth succeeded for 'root' with ke
 
 <!-- lighttpd.
   - Examples:
-  - 2019-10-28 01:39:17: (connections-glue.c.200) SSL: 1 error:1417D18C:SSL routines:tls_process_client_hello:version too low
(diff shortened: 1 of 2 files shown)


ddpbsd:
Thanks for the submission. Here are a few comments.
Let's start the rule IDs with 56201. 

I get this error when loading the rule file:
```
arrakis# cat /tmp/sample_logs | /var/ossec/bin/ossec-logtest 
2019/11/02 18:28:38 ossec-testrule: INFO: Reading local decoder file.
2019/11/02 18:28:38 rules_op: Invalid root element "rule".Only "group" is allowed
2019/11/02 18:28:38 ossec-testrule(1220): ERROR: Error loading the rules: 'lighttpd_rules.xml'.
```
So we need something like this before and after the rule:
```
<group name="local,lighttpd">
   RULE
</group>
```

`<regex>` probably isn't necessary, `<match>` will work just fine.

atomicturtle:
Looks good, thanks again for the contribution!

PR #1794, opened 2019-11-15
Added non-standard Sophos UTM syslog timestamp format to pre-decoding.

mwidman:
The Sophos UTM firewall uses syslog to transmit data but uses a non-standard timestamp format that looks like: 2019:11:06-00:08:03.  Outside of the timestamp, all other aspects appear to be the same as regular syslog entries so it makes sense to me to parse it as such.


ddpbsd:
Thanks for the pull request!

PR #1828, opened 2020-01-21
Replace geoip database support with maxminddb

ddpbsd:
The former geoip databases are no longer maintained, but [libmaxminddb](https://github.com/maxmind/libmaxminddb) is. So switch to this.

Currently the output is limited to the country iso code. I haven't tried very hard to get state/province information, but this can be added if it's desired (I don't see a lot of use in it personally, but I'm not opposed to adding it).

I've only tested with the `GeoLite2-Country.mmdb` database.
The `<geoipdb>` configuration setting has been re-used for this.
If the database file cannot be opened, an error is logged. Should it exit as well?


sempervictus:
They added uniquely identified downloads @ end of last year, any concern with tracking/privacy?

ddpbsd:
@sempervictus Not from me. It's up to the user to consider that.
https://db-ip.com/db/lite.php also has compatible databases (I haven't looked at them yet).

PR #1830, opened 2020-01-25
Update cloudflare-ban.sh

libellux:
Remove the use of cmd jq and use posix

### DIFF
--- a/active-response/cloudflare-ban.sh
+++ b/active-response/cloudflare-ban.sh
@@ -57,7 +57,7 @@ elif [ "x${ACTION}" = "xdelete" ]; then
    -H "X-Auth-Key: $TOKEN" \
    -H "Content-Type: application/json")
    
-   ID=$(echo $JSON | jq -r '.result[].id')
+   ID=$(echo $JSON | grep -Pom 1 '"id": "\K[^"]*' | head -1)
     
    curl -sSX DELETE "https://api.cloudflare.com/client/v4/user/firewall/access_rules/rules/$ID" \
    -H "X-Auth-Email: $USER" \


ddpbsd:
Which release/version/whatever of posix specifies `-P`?
```
[ddp@arrakis:~] :; grep -Pom 1 '"id": "\K[^"]*'
grep: unknown option -- P
usage: grep [-abcEFGHhIiLlnoqRsUVvwxZ] [-A num] [-B num] [-C[num]] [-e pattern]
        [-f file] [-m num] [--binary-files=value] [--context[=num]]
        [--label=name] [--line-buffered] [pattern] [file ...]
```

featzor:
Version: grep (GNU grep) 3.1

```
grep --help
Usage: grep [OPTION]... PATTERN [FILE]...
Search for PATTERN in each FILE.
Example: grep -i 'hello world' menu.h main.c

Pattern selection and interpretation:
  -P, --perl-regexp         PATTERN is a Perl regular expression
```

> grep understands three different versions of regular expression syntax: ``basic'' (BRE), ``extended'' (ERE) and ``perl'' (PRCE).

ddpbsd:
That's the grep version. I don't see `-P` as an option in the [specification](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/grep.html), but maybe I'm looking in the wrong place.

featzor:
Hm you're correct @ddpbsd wonder if this extension of my grep was updated once I installted the PCRE library for OSSEC as I've never intentionally extended grep. Let me have a closer look as well.

ddpbsd:
It's probably a gnu extension. I see it on my linux systems.
I looked at a few other systems quickly (their online manual pages):

- FreeBSD: supported

- NetBSD: not supported

- OpenBSD: not supported

- Solaris: not supported

- AIX: not supported

- MacOS: not supported

The script could check for the `-P` option, and if it isn't found fall back to `jq`. All of this seems more complicated than installing jq, but that's something I generally do anyway.

featzor:
@ddpbsd agreed lets merge this pull request. Could just inform about the dependency requirement in the cloudflare-ban.sh script.

PR #1846, opened 2020-02-27
Add ossec-aws-waf.sh

Midi12:
Add a script to automatically add offenders IP to an existing AWS WAF IP set.


atomicturtle:
This is fantastic!

Issue #1367, opened 2018-02-21
Are there any plans to support journald logs?


q2dg:
It seems it's pretty common nowadays having this kind of logs in current distros (Debian/Ubuntu, Fedora/CentOS, etc) instead of classical syslog ones.
Thanks!

ddpbsd:
Not at the moment. I'm still hoping they see how bad of an idea journald is and revert.

atomicturtle:
I think the biggest limitation with journald right now is that it buffers, so its not a mechanism that lets us analyze data in real time. That definitely puts it at a disadvantage to other kinds of log based intrusion detection

ddpbsd:
Looking at the documentation (*shock*), it looks like the `ForwardToSyslog` option might be do-able. I'll try taking a look at it after I'm done with the current things I'm working on.
https://www.freedesktop.org/software/systemd/man/journald.conf.html#ForwardToSyslog=

sempervictus:
@atomicturtle: that may actually be an advantage if disks went RO for some reason, or space filled up from a zip bomb all of a sudden (though I've not seen one in a while, and no, please don't send me one :-) ).
What about hooking journald itself, or using _journalctl -f_ to track subject units (or all the things)? There is a similar issue with Beats collectors, for which someone wrote journal beat as it required a wholly separate implementation (or ElasticCo folks had the same glimmer of hope for sanity to return).

bbayles:
For what it's worth, modern `journalctls` support a cursor, which allows for easy following along. I've used the `journalctl --show-cursor --unit <whatever>` command and then followed a log with the `--after-cursor` switch before.

dwellkamp:
Is there any progress on this issue? I'd really appreciate it.
Without journald support, logfile monitoring with ossec becomes more and more difficult on almost any modern Linux systems.

ddpbsd:
@dwellkamp I haven't seen any pull requests related to this, and no one has asked questions about implementing anything. The work to support crippled systemdOS is either secret or non-existent. If you need help getting started, write down some ideas of how you'd like to tackle it and we can all look into it.

dwellkamp:
Since I am not a developer/programmer I'd describe what my thought are.

Well, I'm no expert on this topic, but it seems systemd already got an (optional?) HTTP interface to follow it's logs called "systemd-journal-gatewayd".
I found this documentation on how to use it: https://www.freedesktop.org/software/systemd/man/systemd-journal-gatewayd.service.html

[...]

PR #1852, opened 2020-03-05
WIP: journald reader

sempervictus:
This is absolutely not ready to merge, its scaffold for the badly needed systemd journald reader mechanism. So far all it does, hopefully, is open the journal and start reading messages from the end into our queue. Its late, time's short, so this is very rough and untested so far. Collaboration very welcome. 

This seeks to address #1367.

ping @atomicturtle @ddpbsd @dcid

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -65,8 +65,10 @@ OSSEC_LDFLAGS=${LDFLAGS} -lm
 ifneq (${TARGET},winagent)
 ifeq (${uname_S},Linux)
 		DEFINES+=-DINOTIFY_ENABLED
-#		DEFINES+=-DUSE_MAGIC
 		OSSEC_LDFLAGS+=-lpthread
+		DEFINES+=-DHAVE_SYSTEMD
+		OSSEC_LDFLAGS+=-lsystemd
+#		DEFINES+=-DUSE_MAGIC
 #		OSSEC_LDFLAGS+=-lmagic
 else
 ifeq (${uname_S},AIX)
--- a/src/Makefile
+++ b/src/Makefile
@@ -30,6 +30,7 @@ USE_ZEROMQ?=no
 USE_GEOIP?=no
 USE_INOTIFY=no
 USE_PCRE2_JIT=yes
+USE_SYSTEMD?=yes
 
 ifneq (${TARGET},winagent)
 	USE_OPENSSL?=auto
@@ -66,8 +67,10 @@ ifneq (${TARGET},winagent)
 ifeq (${uname_S},Linux)
 		DEFINES+=-DINOTIFY_ENABLED
 		OSSEC_LDFLAGS+=-lpthread
+ifeq (${USE_SYSTEMD},yes)
 		DEFINES+=-DHAVE_SYSTEMD
 		OSSEC_LDFLAGS+=-lsystemd
+endif
 #		DEFINES+=-DUSE_MAGIC
 #		OSSEC_LDFLAGS+=-lmagic
 else
@@ -605,6 +608,7 @@ settings:
 	@echo "    USE_INOTIFY:      ${USE_INOTIFY}"
 	@echo "    USE_SQLITE:       ${USE_SQLITE}"
 	@echo "    USE_PCRE2_JIT:    ${USE_PCRE2_JIT}"
+	@echo "    USE_SYSTEMD:      ${USE_SYSTEMD}"
 	@echo "Mysql settings:"
 	@echo "    includes:         ${MI}"
 	@echo "    libs:             ${ML}"
--- a/src/logcollector/logcollector.c
+++ b/src/logcollector/logcollector.c
@@ -156,6 +156,8 @@ void LogCollectorStart()
 #ifdef HAVE_SYSTEMD
             verbose(READING_JOURNAL, ARGV0, logff[i].file);
             sd_read_journal(logff[i].file);
+#else
+            merror("%s: WARN: journald not available on this version of OSSEC", ARGV0);
 #endif
             logff[i].file = NULL;
             logff[i].command = NULL;
--- a/src/logcollector/logcollector.c
+++ b/src/logcollector/logcollector.c
@@ -152,6 +152,15 @@ void LogCollectorStart()
                        ARGV0);
             }
         }
+        else if (strcmp(logff[i].logformat, "journald") == 0) {
+#ifdef HAVE_SYSTEMD
+            verbose(READING_JOURNAL, ARGV0, logff[i].file);
+            sd_read_journal(logff[i].file);
+#endif
+            logff[i].file = NULL;
+            logff[i].command = NULL;
+            logff[i].fp = NULL;
+        }
 
         else {
             logff[i].command = NULL;
--- a/src/logcollector/logcollector.c
+++ b/src/logcollector/logcollector.c
@@ -80,6 +80,7 @@ void LogCollectorStart()
                 logff[i].file = NULL;
                 logff[i].command = NULL;
                 logff[i].fp = NULL;
+                logff[i].ptr = NULL;
 
                 break;
             }
@@ -99,6 +100,7 @@ void LogCollectorStart()
             logff[i].file = NULL;
             logff[i].command = NULL;
             logff[i].fp = NULL;
+            logff[i].ptr = NULL;
         }
 
         else if (strcmp(logff[i].logformat, "eventchannel") == 0) {
@@ -116,11 +118,13 @@ void LogCollectorStart()
             logff[i].file = NULL;
             logff[i].command = NULL;
             logff[i].fp = NULL;
+            logff[i].ptr = NULL;
         }
 
         else if (strcmp(logff[i].logformat, "command") == 0) {
             logff[i].file = NULL;
             logff[i].fp = NULL;
+            logff[i].ptr = NULL;
             logff[i].size = 0;
 
             if (logff[i].command) {
@@ -138,6 +142,7 @@ void LogCollectorStart()
         } else if (strcmp(logff[i].logformat, "full_command") == 0) {
             logff[i].file = NULL;
             logff[i].fp = NULL;
+            logff[i].ptr = NULL;
             logff[i].size = 0;
             if (logff[i].command) {
                 logff[i].read = read_fullcommand;
@@ -155,13 +160,13 @@ void LogCollectorStart()
         else if (strcmp(logff[i].logformat, "journald") == 0) {
 #ifdef HAVE_SYSTEMD
             verbose(READING_JOURNAL, ARGV0, logff[i].file);
-            sd_read_journal(logff[i].file);
+            logff[i].read = sd_read_journal;
 #else
             merror("%s: WARN: journald not available on this version of OSSEC", ARGV0);
 #endif
-            logff[i].file = NULL;
             logff[i].command = NULL;
             logff[i].fp = NULL;
+            logff[i].ptr = NULL;
         }
 
         else {
@@ -206,6 +211,7 @@ void LogCollectorStart()
                     if (logff[i].fp) {
                         fclose(logff[i].fp);
                         logff[i].fp = NULL;
+                        logff[i].ptr = NULL;
                     }
                     logff[i].file = NULL;
                 }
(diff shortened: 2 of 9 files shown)


src/Makefile
```diff
@@ -65,8 +65,10 @@ OSSEC_LDFLAGS=${LDFLAGS} -lm
 ifneq (${TARGET},winagent)
 ifeq (${uname_S},Linux)
 		DEFINES+=-DINOTIFY_ENABLED
-#		DEFINES+=-DUSE_MAGIC
 		OSSEC_LDFLAGS+=-lpthread
+		DEFINES+=-DHAVE_SYSTEMD
```
ddpbsd: Should this be conditional somehow for older linux systems?

src/logcollector/read_journald.c
```diff
@@ -0,0 +1,64 @@
...
+#include "logcollector.h"
+#include <systemd/sd-journal.h>
+
+void prime_sd_journal(sd_journal **jrn) {
+  int ret;
+  ret = sd_journal_open(&jrn, SD_JOURNAL_LOCAL_ONLY);
+  ret = sd_journal_get_fd(jrn);
+  ret = sd_journal_seek_tail(jrn);
+  // prime sd_journal_next before the reader loop
+  ret = sd_journal_previous(jrn);
```
ddpbsd: `ret` should be checked for failure.

src/logcollector/logcollector.c
```diff
@@ -152,6 +152,17 @@ void LogCollectorStart()
                        ARGV0);
             }
         }
+        else if (strcmp(logff[i].logformat, "journald") == 0) {
+#ifdef HAVE_SYSTEMD
+            verbose(READING_JOURNAL, ARGV0, logff[i].file);
```
ddpbsd: `READING_JOURNAL` needs to be declared somewhere (`src/error_messages/error_messages.h`?)

ddpbsd:
Thanks for jumping on this!
I'm sure some of the comments I'm leaving are already known issues, but my memory is bad enough that I've learned to leave myself notes.

src/logcollector/read_journald.c
ddpbsd: Some of these will also complain about the pointer. They expect `*` but get `**`.

src/logcollector/read_journald.c
```diff
@@ -0,0 +1,64 @@
...
+  const char *data;
+  size_t len;
+  struct timeval tv;
+  uint64_t curr_timestamp;
+  time_t nowtime;
+  struct tm *nowtm;
+  char tmbuf[64], final_msg[OS_MAXSTR];
+
+  prime_sd_journal(&jrn);
+  ret = sd_journal_next(j);
```
ddpbsd: Should be `et = sd_journal_next(jrn);`

src/logcollector/read_journald.c
```diff
@@ -0,0 +1,64 @@
...
+  // Anything negative is an error
+  while (ret >= 0) {
+    // No messages, wait for data
+    if (ret == 0) {
+      while ((sd_journal_wait(jrn, (uint64_t)-1)) == SD_JOURNAL_NOP) {
+        sleep(1);
+      }
+    } else {
+      // Read data
+      ret = sd_journal_get_data(jrn, "MESSAGE", (const void **)&data, &len);
```
ddpbsd: `MESSAGE` doesn't include the hostname. Not sure if the log passed into the queue later matches what OSSEC is expecting. When I played with this stuff previously, I also grabbed `_HOSTNAME` and `SYSLOG_IDENTIFIER`

src/logcollector/read_journald.c
sempervictus: This is why professional developers sleep before coding, ty, will fix.

src/logcollector/read_journald.c
sempervictus: Thank you, very useful insight, will play around with these - if you get a chance to do a compile/run test, would love to get feedback on which ones go where.

sempervictus:
Nice! It works - was worried merge-squash would lose authorship.
Thanks @illuusio.

sempervictus:
Prototyped a source selector for the journal reader which should allow us to filter which units (now SYSLOG_IDENTIFIERs) we're reading from. This is the "cheapest" way i could think of to do this in a situation where we may have tons of subscribers to individual journals, doing that check and failing 99% of the time. Not sure what that'll do if nothing's passed in, but in my little console version of this it does perform the filtering correctly and writes out the unit name. Reason for `strstr` vs direct comparison is that systemd can generate auto-named units with common prefixes, and this should help pull from them even if they're added after the reader is initialized.

@ddpbsd: will add hostname too, thanks as always for the useful suggestions.

@atomicturtle, could you please poke at that `snprintf` to make the message format OSSEC-friendly as a syslog message or whatever format you'd prefer it to have in the queue?

src/Makefile
sempervictus: Yes, yes it should. My makefile-fu is weak, barely grasshopper status, but IIUC this would happen when configure/autogen determine if the systemd headers are available to compile against. Wanna toss me a PR? :)

src/logcollector/read_journald.c
sempervictus: Both _HOSTNAME and SYSLOG_IDENTIFIER are now in there for prodding and reorg, but i think i have all the data requested...

src/Makefile
ddpbsd: There's no configure/autogen. I'll try to get you a diff soon.

ddpbsd: I started a PR, but it looks like your tree is older. So the PR would be huge with a bunch of stuff unrelated to this. It's the last few commits [here](https://github.com/ddpbsd/ossec-hids/tree/reader_journald). If there's a better way to get you a diff, let me know.

sempervictus: Updated my GH master, so now my branch is equivalent to this PR and yours should apply cleanly. Thank you

sempervictus: Solved - cherry-picked and squashed your commits locally, rebased on master, force-pushed up. Thank you sir.

sempervictus:
Awesome, looks like Dan fixed the CI.

So who here is comfortable enough with positional cursors on source data to help figure out how we want to deal with keeping track of where we were for each reader before a restart or other feed interrupting event?

sempervictus:
@atomicturtle @ddpbsd @dcid: i think we have hit a small architectural pain point here: our _reader_ and _parser_ code is intertwined, and now that audit logs can come from journald (file or memory) or a flat file (fd), the fact that the audit reader is its own encapsulated file, hard-coded to the fd approach, either means i copy-paste code (same for postgres and others) and make a ton of redundant bloat, or we think about splitting out actual accessors of API/ABI endpoints and file descriptors (the literal "readers") from the parsing and formatting code.
Thoughts? Interns to sacrifice on the altar of clean-code?

src/logcollector/logcollector.c
```diff
@@ -152,6 +152,17 @@ void LogCollectorStart()
...
+            sd_read_journal(logff[i].file);
```
illuusio: Just wondering is this correct way of handling this as other reader just make logff[i].read = sd_read_something(with args) and journald reader hangs forever?

src/logcollector/read_journald.c
```diff
@@ -0,0 +1,106 @@
...
+      // Check if data is coming from the unit starting with our passed selector
+      ret = sd_journal_get_data(
+        jrn,
+        "SYSLOG_IDENTIFIER",
+        (const void **)&jsrc,
+        &len
+      );
+      // Strip off "SYSLOG_IDENTIFIER=" prefix for unit comparison inline
+      if (strstr((char *)(jsrc + 18), unit) == (char *)(jsrc + 18) ) {
+        // Read hostname
```
illuusio: There should be a way to log all journald stuff as reading /var/log/messages as this just logs some units output.

ddpbsd:
@sempervictus It depends on how quickly you want journald support. Adding another reader in the same way previous ones were added should be eaasier than refactoring all/most of them.

sempervictus:
@ddpbsd: makes sense.
@illuusio: thanks for all the commits, merged.
@all: so cursors are the remaining item on the list - we need to be able to pick up where we left off during an agent restart, ideally.

illuusio:
@sempervictus only thing with cursor is where to store them? If used multiple inputs is there some commong way in OSSEC-HIDS to store such info if there is not we shoud re-invent wheel with JSON or such?

illuusio:
Now this should be Beta quality as I got it crashing couple time but I we can get some testers (I have been testing this on couple servers and it work quit nice.

sempervictus:
@illuusio: what're you using for your decoders and rules to see data, or are you seeing data hit the manager?

illuusio:
@sempervictus I just hardcoded version that printed on analysid inner things. Servers sends me emails about something is wrong. I have to investigate more to see if rules really work as they should but messages are parsed (and separated) as they are coming from syslog-ng (which it is mimicking).

illuusio:
I've been testing this while now in productions and it works fairly well. It needs to have some rules for journald but I think this could be merged and rest of the problems solved after it receives more testing from community.

sempervictus:
Yeah, i'm good with pushing this into master and wider testing.
One remaining through comes to mind: do we want to permit the reader to restrict which _hostname_ it reads from? There are two competing and completely valid use cases here:
1. An OSSEC-equipped host logging all of its dockers or what not
2. An OSSEC-equipped host with OSSEC-equipped LXD/LXC containers _belonging to someone else_ - like a cloud provider who doesn't want their tenants data for any number of reasons.

sempervictus:
@illuusio: this effort is going to be the subject of a talk or two, any objections to me giving you a shout-out for all the hard work by the name listed in your GH profile? Any prefered alternative for given name or handle than the ones listed there?

illuusio:
>     1. An OSSEC-equipped host logging all of its dockers or what not
> 
>     2. An OSSEC-equipped host with OSSEC-equipped LXD/LXC containers. For all we know, they have separate OSSEC masters, and only want their own data.

Is there such restrictions from other readers? I'm just asking as this kind of a easy task but should be done as 'OSSEC'-way if there is any and is there some <xml>-tag in config for this?

> Any prefered alternative for given name or handle than the ones listed there?

Nope just use that if you prefer to give credit but is not necessary as part of security in your life is to play low :smile_cat:

sempervictus:
@illuusio: There are no such restrictions from other readers AFAIK, but other readers access inodes which are inside the namespace of the container, whereas the host's journal will have entries from journald-aware child containers.
First talk on this is today @ 1530 EST in the OSSEC virtual webinar (COVID replacement for the normal annual conference)

illuusio:
Hopefully there is some video as it's kinda late for me. So If I understood correctly the problem is that you have bunch of containers which report to one server and there should be restricting tag to make it choose which it allows to take input like:
`<allow-host>megahost</allow-host>
 <allow-host>container-host1</allow-host>`
Or I'm missing something.

sempervictus:
@illuusio: we'll have video, either atomicorp will have a copy or both they and i will.
Since the configurations have to apply to _all_ OSSEC agents, i'd say we do something like:
```
<journal-namespace>local</journal-namespace>
```
or
```
<journal-namespace>all</journal-namespace>
```
and then have it pull the current NS' hostname to compare

illuusio:
Ok and local only reports local stuff and all everything. I take a look as I have to get OSSEC working with agents soon and need this one be number one citizen.

ddpbsd:
Please add `libsystemd-dev` to the list of required packages in INSTALL.

sempervictus:
@illuusio: thank you. Based on the conference today, sounds like we want to be able to filter host/host+containers, which log source (application/service), and even potentially a substring match in the event output, though i think that can be added later - we just wanna hit master branch with a stable configuration API (in terms of not changing parameters down the line, only potentially adding some if people want more filtering capacity @ the reader).
@ddpbsd: good catch! wouldn't have caught that since Arch has the development libraries right in the main `-libs` package

atomicturtle:
merging this in so we can start more open testing

bntuser:
what do i have to write in ossec.conf to start reading the output of journalctl

PR #1875, opened 2020-05-23
Shuffle some variables around

### LINKED ISSUE #1871: Build fails with GCC-10 and GCC-9 when -fno-common option is used
GCC version 10 will soon arrive, and one of the planned changes is switching from `-fcommon` (default in GCC 9) to the new default `-fno-common`.

As documented in Gentoo Linux [issue 720088](https://bugs.gentoo.org/720088), building OSSEC 3.6.0 fails with typical linker errors when `-fno-common` is used:

```
/usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/../../../../x86_64-pc-linux-gnu/bin/ld: os_net.a(os_net.o):
(.bss+0x0): multiple definition of `agt'; client-agent/config.o:(.bss+0x0): first defined here
collect2: error: ld returned 1 exit status
make: *** [Makefile:1016: ossec-agentd] Error 1
```

We have a Gentoo [Wiki article](https://wiki.gentoo.org/wiki/Gcc_10_porting_notes/fno_common) describing the problem in more detail. While patching as a temporary measure is possible (I am involved in my role as the designated Gentoo package maintainer), the OSSEC source code needs to be fixed.

ddpbsd:
Fixes #1871


ddpbsd:
Updated with a successful server build.
Tested building with mysql, postgresql, and geoip enabled/disabled.

ddpbsd:
Updated to current HEAD. Compile tested only.

atomicturtle:
Tested out fine here too, merging this in

PR #1880, opened 2020-06-03
Fix IPv4 and IPv6 issue where OSSEC-HIDS only binds on first IPv6

illuusio:
I don't know if this is correct way to fix this but I couldn't bind OSSEC remoted any other than :: or ::192.168.0.1 if these defines where available. Without them everything works as expected.

### DIFF
--- a/src/os_net/os_net.c
+++ b/src/os_net/os_net.c
@@ -64,32 +64,6 @@ OSNetInfo *OS_Bindport(char *_port, unsigned int _proto, const char *_ip)
     /* init hints for getaddrinfo() */
     memset(&hints, 0, sizeof(struct addrinfo));
 
-   /*
-    * If you cannot bind both IPv4 and IPv6, the problem is likely due to the
-    * AF_INET6 family with the AI_V4MAPPED flag. Alter your Makefile to use the
-    * NOV4MAP define and it should work like a breeze. All of the *BSDs fall
-    * into this category even though AI_V4MAPPED exists in netdb.h (true for
-    * all modern OS's). This should work with all Linux versions too, but the
-    * original code for AF_INET6 was left for Linux because it works.
-    *
-    * d. stoddard - 4/19/2018
-    */
-
-#if defined(__linux__) && !defined(NOV4MAP)
-#if defined (AI_V4MAPPED)
-    hints.ai_family = AF_INET6;		/* Allow IPv4 and IPv6 */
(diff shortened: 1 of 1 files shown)


ddpbsd:
What is the actual problem you're trying to solve?
How have you tested this?

illuusio:
@ddpbsd Yes I have tested it in development openSUSE linux server. OSSEC server standalone works as charm and I like to it as OSSEC remoted host.
If that patch is not applied OSSEC remoted only sees and binds to IPv6 :: (localhost) (not IPv4 127.0.0.1) UDP port and nothing else. If remoted is launch without <local_ip> in <remote> (ossec.conf). If local_ip is used like this in ossec.conf (build with hybrid)
```
<remote>
  <local_ip>192.168.2.1</local_ip>
</remote>
```
Which should bind it to IPv4 ip 192.168.2.1 but it binds to IPv6 ::192.168.2.1 which is not completely wrong but not what should happen.
With patch (This is devel package I use to test https://build.opensuse.org/package/show/home:illuusio/ossec-hids and it's openSUSE) I got without <local_ip> both IPv6 :: and 127.0.0.1 binding which is still wrong as documentation says it should bind to every interface IP and with <local_ip> it binds happily to correct IP and UDP port.
This can be just documentation TL;DR; problem but without that code everything works and with it remoted (I can't use IPv6) everything is bonkers. Probably this more device detection code but as comment says this is left to Linux as it seems to work and when I studied it should not be incorrect but somehow it just not working.

ddpbsd:
Are agents unable to connect? Have you tried defining `NOV4MAP` in the Makefile?
I haven't tried Suse in a while, but it worked last time I tried it. Ubuntu and Centos are able to work with everything at default as well.

illuusio:
No I haven't I test but I try it and report back. Also I haven't got agent working it just says 'ERROR: Unknown imsg type' and I think I just haven't read enough manual. OSSEC-HIDS server works very well.

ddpbsd:
I just did a fresh checkout of master on a fresh opensuse instance and was not able to reproduce the error. I'll have to fiddle with it a bit to see if I can find any settings that cause issues I guess.

illuusio:
Hmm.. did you compile package from source or used rpm from repo? I have to dig this little bit further what is the main reason of this.

ddpbsd:
I installed from source. "Start with what you know." I'll try looking at the package when I can make some time, but I'm not very experienced with Suse.

atomicturtle:
Re-implemented this in PR #2304

PR #1889, opened 2020-06-24
A few small fixes for problems found with Clang analyzer

ddpbsd:
There are plenty more in the report.


atomicturtle:
aaaaaaand I timed this one wrong.. ended up making a conflict by merging that os_dns revert from before

PR #1913, opened 2020-09-23
Fix regexp import in ossec-batch-manager.pl

seren:
Hi, this fixes what appears to be a syntax error in the user-contributed file `ossec-batch-manager.pl`. Currently it produces the error:

```
Global symbol "%RE" requires explicit package name (did you forget to declare "my %RE"?) at ./ossec-batch-manager.pl line 105.
Global symbol "%RE" requires explicit package name (did you forget to declare "my %RE"?) at ./ossec-batch-manager.pl line 107.
BEGIN not safe after errors--compilation aborted at ./ossec-batch-manager.pl line 139.
```

Fixes https://github.com/ossec/ossec-hids/issues/1914


atomicturtle:
Excellent, thanks for the commit!

PR #1919, opened 2020-10-03
Fix NetBSD build.  Without this analysisd can't understand regex.

giannidaprile:
Borrowing from OpenBSD block.  Fixes build on NetBSD.  Without this analysisd can't understand regex.


atomicturtle:
Thanks for the commit!  For what its worth, this is a performance setting.

PR #1967, opened 2021-04-12
Update ssh_asa-fwsmconfig_diff

tonoitp:
Added en expect at line 174 in case a cisco device does not require enable statement


atomicturtle:
Excellent catch!

PR #1974, opened 2021-05-21
[Documentation] Updating Broken Documentation Links

lyellread:
Fixing the broken links that result from abandoning http://www.ossec.net/en/manual.html and moving to http://www.ossec.net/docs/index.html. 

I have made my best attempt to match up the appropriate page in the new documentation with where I think the old URL pointed. 

Please feel free to edit this if you see something off.

~ Lyell


ddpbsd:
Wow, thanks for going through all that!

lyellread:
> Wow, thanks for going through all that!

No issues at all. I missed a couple (ones I could not find the replacements for), I documented them in #1975.

PR #2029, opened 2021-12-26
active response script for nftables

ChristianBeer:
Does the same as firewall-drop.sh and firewalld-drop.sh but for nftables (default in Debian 10+). Needs a bit more steps to setup but does a very good job.
Documentation should be updated to contain maybe an example for the nftables configuration.

Tested on Debian 11 with IPv4 and IPv6 adresses.

### DIFF
--- a/active-response/nftables-drop.sh
+++ b/active-response/nftables-drop.sh
@@ -1,0 +1,185 @@
+#!/bin/sh
+# Adds an IP to a nftables set
+# Requirements: Linux with nftables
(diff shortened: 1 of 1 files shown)


atomicturtle:
Thanks for the PR, this is great work!

ChristianBeer:
@atomicturtle Thanks for including this in the 3.7.0 release. Don't forget to merge the documentation PR too as this contains a useful config snippet.

Issue #2020, opened 2021-10-24
False positive "Trojaned version of file '/bin/diff' detected" on Archlinux


tiiiecherle:
Hey Ossec Team,

with the latest version diffutils 3.8-1 installed ossec reports a trojaned version of a few files.

```
OSSEC HIDS Notification.
2021 Oct 22 10:22:02

Received From: archvbox->rootcheck
Rule: 510 fired (level 7) -> "Host-based anomaly detection event (rootcheck)."
Portion of the log(s):

Trojaned version of file '/bin/diff' detected. Signature used: 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' (Generic).



--END OF NOTIFICATION



OSSEC HIDS Notification.
2021 Oct 22 10:22:02

Received From: archvbox->rootcheck
Rule: 510 fired (level 7) -> "Host-based anomaly detection event (rootcheck)."
Portion of the log(s):

Trojaned version of file '/sbin/diff' detected. Signature used: 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' (Generic).



--END OF NOTIFICATION



OSSEC HIDS Notification.
2021 Oct 22 10:22:02

Received From: archvbox->rootcheck
Rule: 510 fired (level 7) -> "Host-based anomaly detection event (rootcheck)."
Portion of the log(s):

Trojaned version of file '/usr/bin/diff' detected. Signature used: 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' (Generic).



--END OF NOTIFICATION



OSSEC HIDS Notification.
2021 Oct 22 10:22:02

Received From: archvbox->rootcheck
Rule: 510 fired (level 7) -> "Host-based anomaly detection event (rootcheck)."
Portion of the log(s):

Trojaned version of file '/usr/sbin/diff' detected. Signature used: 'bash|^/bin/sh|file\.h|proc\.h|/dev/[^n]|^/bin/.*sh' (Generic).
```

I opened an issue at the archlinux bug tracker here:
[https://bugs.archlinux.org/task/72519#comment203202](https://bugs.archlinux.org/task/72519#comment203202)

When testing the files against virustotal database nothing suspicious is reported and the checksum seems fine.

Changing the diff line in rootkit_trojans.txt to `diff !bash|^/bin/sh|file\.h|proc\.h|^/bin/.*sh!` solves the reporting.

[...]

PR #2062, opened 2022-09-07
Fix issue #2020

ddpbsd:
/bin/diff returns /dev/full on fedora, so remove the /dev check


sempervictus:
I've been using
```xml
  <rule id="730004" level="4">
    <if_sid>510</if_sid>
    <field name="file">bin/mail$|bin/diff$</field>
    <description>False-positive match for rootcheck regex</description>
  </rule>
  ```
  for this lately - spams on Arch like you wouldn't believe :)

PR #2105, opened 2023-11-24
* insert into db true NULLs instead string "NULL"

AdUser:
This patch changes insertion of srcip/dstip alert fields to database table from "NULL" (varchar string) value to true sql NULLs.
it takes less space in table and add ability to use IS NULL/NOT NULL/... sql expressions instead slower string comparision.

```
CREATE TABLE alert
    (
   ...
    src_ip          VARCHAR(46), -- nullable
    dst_ip          VARCHAR(46), -- nullable
    ...
    );
```


AdUser:
P.S. This patch tested with postgresql database about ten months, and have no issues in my ossec setup, but feel free to do more testing if you consider so.

PR #2125, opened 2024-05-23
Code  base Improvement 

JohnKimaiyo:
To improve the given Base64 encoding and decoding code, consider the following enhancements:

Reduce redundancy and simplify conditions.
Use consistent and meaningful naming conventions.
Add error handling and parameter validation.
Optimize memory allocation.

### DIFF
--- a/src/addagent/b64.c
+++ b/src/addagent/b64.c
@@ -24,202 +24,102 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-
-#define TRUE    1
-#define FALSE   0
+#include <stdbool.h>  // For using bool instead of int for boolean functions
 
 /* Prototypes */
-static int is_base64(char c);
-static char encode(unsigned char u);
-static unsigned char decode(char c);
+static bool is_base64(char c);
+static char encode_char(unsigned char u);
+static unsigned char decode_char(char c);
 
-/* Global variables */
+/* Global functions */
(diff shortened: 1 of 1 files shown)


JohnKimaiyo:
To improve the given Base64 encoding and decoding code, consider the following enhancements:

Reduce redundancy and simplify conditions.
Use consistent and meaningful naming conventions.
Add error handling and parameter validation.
Optimize memory allocation.

JohnKimaiyo:
To improve the given Base64 encoding and decoding code, consider the following enhancements:

Reduce redundancy and simplify conditions.
Use consistent and meaningful naming conventions.
Add error handling and parameter validation.
Optimize memory allocation.

atomicturtle:
Thanks for the PR, @JohnKimaiyo.

We're going to close this without merging. The change is a style rewrite of the Base64 helpers used for agent key import/export, with no accompanying bug fix or tests. The rewritten decode path also introduces real issues relative to what's on master today (out-of-bounds reads on padding checks when the filtered length isn't a multiple of 4, and encode_base64's const signature wasn't updated in the headers).

The current monit-derived b64.c is long-standing and working for that path, so the risk of a wholesale restyle here outweighs the benefit. If there's a concrete Base64 bug to fix later, a small targeted PR with bounds-safe decode and header updates would be the right shape.

Appreciate the interest in cleaning this up.

PR #2158, opened 2025-06-20
Create cmake-multi-platform.yml

cmac9203:


atomicturtle:
starting some work on here, thanks for the idea

Issue #447, opened 2014-11-22
ossec-dbd not reconnecting after database failure


ChristianBeer:
I was babysitting a new server and ran into memory trouble twice. During this I noticed an odd behaviour of ossec-dbd. The mysql server threads where killed at 03:57:53 (out of memory) but in the same second restarted. Ossec looses connectivity for only this query but can't reconnect after this. I have to manually restart ossec in order to reconnect ossec-dbd back to the database.

Can someone else please confirm this behaviour? Do I have to change something on the mysql server side?

```
2014/11/22 03:57:58 ossec-dbd(5203): ERROR: Error executing query 'SELECT id FROM location WHERE name = 'hopper->/var/log/messages' AND server_id = '1' LIMIT 1'. Error: 'MySQL server has gone away'.
2014/11/22 03:57:58 ossec-dbd(5209): INFO: Closing connection to database.
2014/11/22 03:57:58 ossec-dbd(5210): INFO: Attempting to reconnect to database.
2014/11/22 03:57:58 ossec-dbd(5202): ERROR: Error connecting to database 'localhost'(ossec): ERROR: Unknown MySQL server host 'localhost' (2).
2014/11/22 03:58:00 ossec-dbd(5210): INFO: Attempting to reconnect to database.
2014/11/22 03:58:00 ossec-dbd(5202): ERROR: Error connecting to database 'localhost'(ossec): ERROR: Unknown MySQL server host 'localhost' (2).
2014/11/22 03:58:04 ossec-dbd(5210): INFO: Attempting to reconnect to database.
2014/11/22 03:58:04 ossec-dbd(5202): ERROR: Error connecting to database 'localhost'(ossec): ERROR: Unknown MySQL server host 'localhost' (2).
2014/11/22 03:58:12 ossec-dbd(5210): INFO: Attempting to reconnect to database.
2014/11/22 03:58:12 ossec-dbd(5202): ERROR: Error connecting to database 'localhost'(ossec): ERROR: Unknown MySQL server host 'localhost' (2).
2014/11/22 03:58:28 ossec-dbd(5210): INFO: Attempting to reconnect to database.
2014/11/22 03:58:28 ossec-dbd(5202): ERROR: Error connecting to database 'localhost'(ossec): ERROR: Unknown MySQL server host 'localhost' (2).
2014/11/22 03:59:00 ossec-dbd(5210): INFO: Attempting to reconnect to database.
[...]

PR #2184, opened 2026-02-08
Ossec-dbd connector fix

atomicturtle:
ossec-dbd will reset its error_count upon successful reconnection, preventing it from incorrectly hitting the error limit due to accumulated transient failures.


atomicturtle:
closes issue #447

Issue #2208, opened 2026-04-15
Consider including detected file hash in full_log alert output


Hu6li:
When analyzing alerts generated by OSSEC, I noticed that the full_log output does not seem to include the cryptographic hash (e.g., SHA256 or MD5) of the detected files. 

Would it make sense to add this information to the full_log output in alerts?

My reasoning:
- Having the hash directly in the alert would help with quick investigation and correlation with threat intelligence databases.
- It would allow for easier identification of specific files flagged by multiple alerts and help track changes over time.

Is there an architectural or performance tradeoff to consider here? What are your thoughts on adding the detected file hash to the alert full_log?

Personally, I think this addition could improve incident investigation and response workflows.
If you consider it a valuable addition, I am happy to open a PR.

atomicturtle:
So along those lines I want to extend the rule language to allow you to call those variables from the decoded event in the <description> field for an event. Like this <description>File changed $(syscheck.path) Hash $(hash.sha256)</description>.  This would be especially powerful since we have the capability to dynamically declare variables with the dynamic decoder engine. Wanna take a shot at that?

Hu6li:
Sorry for my late response, I will try and give it a go 👍 
I first have to look into the dynamic decoder engine as I currently don't know how this works.

janderik:
Great suggestion! Including file hashes in alert output would be very valuable for several reasons:

1. **Threat Intelligence Correlation**: Having the hash directly in the alert enables immediate lookup in VirusTotal, MISP, or other TI platforms without manual extraction.

2. **Incident Response Efficiency**: IR teams can quickly pivot from alert to investigation without extra steps.

3. **Alert Deduplication**: Hashes help identify when multiple alerts reference the same file across different hosts.

**Implementation considerations:**

- **Performance**: Hash computation (especially SHA256) on every file change could impact performance on high-traffic systems. Consider making it configurable or only computing hashes when specifically enabled.

- **Storage**: Hashes add a few bytes per alert, but this is negligible compared to the full_log itself.

- **Configuration option**: Something like `<alerts>hash_in_full_log>yes</alerts>` would let users opt-in based on their needs.

- **Hash type preference**: MD5 is faster but weaker; SHA256 is more secure but slightly slower. Could offer both or let users choose.

**Alternative approach**: Instead of modifying full_log, you could add a new field like `file_hash` to the alert JSON structure. This would be cleaner for parsing and wouldn't break existing log parsers.

[...]

PR #2244, opened 2026-07-24
rootcheck: append MD5/SHA1/SHA256 hashes of detected files to alerts

Hu6li:
Detected files in check_rc_files, check_rc_trojans and check_rc_sys now have their hashes computed and appended to the alert message. DecodeRootcheck extracts them into Eventinfo for structured JSON output.

This PR adds hash support mentioned in #2208 
As hashing does not add much overhead especially when only triggered after a rootkit detection, I did not implement a config switch.

Please let me know what you think.

### DIFF
--- a/src/analysisd/decoders/rootcheck.c
+++ b/src/analysisd/decoders/rootcheck.c
@@ -112,6 +112,52 @@ static FILE *RK_File(const char *agent, int *agent_id)
     return (NULL);
 }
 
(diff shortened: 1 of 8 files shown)


Hu6li:
did the rework and applied copilots suggestions

atomicturtle:
This was such a great idea. Merged!

Issue #1793, opened 2019-11-13
Flood of "File added to the system" alerts after new agent install


randallsevery:
Every time I add a new agent to my OSSEC installation (version 3.3.0 on CentOS 7 servers), I get a flood of "File added to the system" alerts, even though the files are all of various ages and have definitely not been updated recently.  According to all of the research I did on this issue online, I am supposed to be able to reset the baseline scan for a particular agent by issuing the following command on the Manager server:

/var/ossec/bin/syscheck_control -u AGENT_ID

But even after doing that, and for a while after the agent has been installed, I continue to get new file alerts for already existing files on the agent server.

What is the recommended way to install and configure a new agent to avoid getting thousands of new file alerts in the process?

ddpbsd:
Do you have the alert on new files setting enabled?

randallsevery:
Yes, I have the set the config option "<alert_new_files>yes</alert_new_files>", since the primary purpose of this OSSEC installation is to detect new files that have been added by malware on the servers.  But I only want alerts for new files that have been added *after* the OSSEC agent has been installed, not thousands of alerts for all of the files that have been on the server for months or years.

atomicturtle:
There are a few different ways to handle that. The condition here is that you're adding an new agent and the database for that system is empty at that point. So obviously every file is a "new" file in that condition, since the database for that agent is empty. You could do a things here:

1. Disable alert_new_files  and wait for the database on the new agent to finish generation, then re-enable it
2. Create a rule that ignores new files from just that agent,  and then turn that rule off when the syscheck db is complete

randallsevery:
Thank you for the quick response.  I wonder if the syscheck_control command that is discussed as the way of resetting the baseline didn't do anything *because* the database on the new agent didn't finish generation.  I'll try the option of disabling the alert_new_files option temporarily.  What is the best way to determine that the database on the new agent has finished generation?  It seems like it can take a while, at least from the timing of the alerts that I've been getting.  The main problem I've had with the flood of alerts for new files is that it seems to overwhelm the ossec-maild process, which crashes, and unless I notice that it's no longer running I won't get *any* alerts for a while.

ddpbsd:
The agents log something when they're done with a syscheck run.
Some of my syscheck logs look like this:
```
2019/11/29 15:14:15 ossec-analysisd: SyscheckInit completed.
2019/11/29 15:14:24 ossec-syscheckd: INFO: Starting syscheck scan (forwarding database).
[...]

Issue #1681, opened 2019-03-08
syscheck and OS patching


liedekef:
In the FAQ I can find this:

> How do I stop syscheck alerts during system updates?
> There is no easy way to do this, but there are work-arounds. Stop the OSSEC processes on the manager, and run /var/ossec/bin/syscheck_control -u AGENT_ID. This will clear the syscheck database for the agent, and the next time syscheck runs it will create a new baseline. Next, start the OSSEC processes on the manager. Once the system update is complete, run a syscheck scan on that agent. The database will be populated with new values, and should not trigger “file modified” alarms.

However: why should I need to stop/start the OSSEC processes on the manager? Should this not be stopped on the agent?

ddpbsd:
In an agent/server setup, the syscheck database is kept on the server. Stopping the processes on the agent shouldn't be necessary since the comparison happens on the server itself. I don't think stopping the server processes should be necessary, but it probably ensures that the database isn't open when it's cleared.

liedekef:
But when in agent/server setup, I e.g. imagine having hundreds of servers. I can't just shut down the server process, since that would impact all others too, not just the one or two or ten I plan to update that night (for example).
Wouldn't a more correct action sequence be: stop syscheck on the agent, clear syscheck db on the master, update the agent os, start syscheck on the agent ?
Also, approving a change seems unclear to me. If e.g. one file changes, do I need to approve the change, or does ossec only send out 1 alarm (if configured to do so) and that's it?

ddpbsd:
You could do it the way you propose. If the syscheck db is open on the server, I'm not entirely sure what will happen. It's a sub-optimal solution, but it's what we have at the moment.
I'm not sure what you mean by approving changes.

liedekef:
"Approve a change" means to approve a file integrity change (it is like saying "I approve of the new content"). But maybe ossec just doesn't work like that and just does the integrity calculation+check each time a file gets updated. It would be interesting to know the process of the file integrity workflow.

ddpbsd:
There's no approval. OSSEC simply alerts when a value has changed.
To simplify the explanation of how the whole process works, I'll just use a standalone installation. Agent/server installs have a few more steps (but those are largely unrelated to syscheck itself).

- `ossec-syscheckd` will do an initial scan of a system, as determined by the configuration in `ossec.conf`. 
- `syscheckd` sends each entry (including hashes, mtime, owner, group, and size) to `ossec-analysisd`.
[...]

Issue #1289, opened 2017-10-18
syscheck logs even after restart sometimes


eduda:
I'm not sure if this is a bug or by design, what i want to do is not have syscheck alert on a file change if I know there are changes taking place, for example, yum update (390 rpms installed as part of os update).
what i thought you could do is just stop ossec, update, then start it back up. most of the time, it will not alert on the changes, but in some cases, i've noticed it does alert after the restart on some file, and some times. is that a bug? 

Is there a better way to suppress alerting without stopping ossec?

Thanks.

ddpbsd:
I think the md5 whitelist feature is in master. Other than that, not much you can do.

eduda:
Is it possible that on startup for some reason the database is not updated before it starts alerting? so in some cases, there's a race?

ddpbsd:
I'm not sure what you mean. If an entry in the database gets updated, there should generally be an alert.
If you stop OSSEC, update packages, and stat OSSEC; a syscheck scan should run. Any changes it detects should probably be alerted on, including those from the package updates.

eduda:
So what i need to happen is this. stop ossec, update some rpms, start ossec, but don't have it alert on the rpm update. normally this works fine, but sometimes, we get an alert. i'm not sure why.

ddpbsd:
If the MD5 whitelisting is in MASTER, you can populate the sqlite3 database with the correct hashes and file names. You can also clear the database before running the updates so everything that gets added is new.

eduda:
is there a way to stop osec, make the rpm updates, then tell ossec update the db with any changes before starting realtime monitoring alerting?

ddpbsd:
Not out of the box. If you can figure out how to get the hashes of the files, scripting a solution wouldn't be the worst.

eduda:
The part that i still don't understand is why does it alert sometimes in syscheck realtime and other times it doesn't? 

in this example: stop ossec, update rpms, start ossec.

ddpbsd:
If you have not turned off auto_ignore, that could affect things. Other than that, I don't know.

Issue #677, opened 2015-10-05
syscheck/rootcheck during OS patching


aremai:
as I understand ossec's syscheck module regularly performs file integrity checks on files and compares the current hash with a known good hash for the file. if the hash doesn't match then ossec repots a change to the file....
the database of known good hashes is initally created the first time syscheck runs after ossec (agent/master) is installed. so far so good.

what happens after applying OS patches to the system? when you upgrade your system a lot of packages, e.g binaries will be updated and ergo their hashes will also change.
how to distinguish between a normal OS upgrade and a rootkit?

what is the recommended procedure during OS patching/upgrading?
how to avoid hundreds of alerts? would you just simply stop OSSEC?

mstarks01:
Yes, you'll get a lot of alerts unless you tune them out. But this isn't a defect. Please try your question on the mailing list. Others have come up with creative solutions for this problem.

aremai:
Hi Michael,

I've already tried the mailing list.... @ddpbsd said "There's been some thoughts on how to handle this, but no conclusions and no actions as of yet."

I thought there was more of a "general way" to avoid getting flooded with alerts during patching. Something like a best-practice or so.

I think the main issue during patching (correct me if I'm wrong) is the file integrity checks, because obviously the hashes change after patching. Log analysis wouldn't be such an issue I suppose.

mstarks01:
IMO, the right way to do this would be to decode hashes from the syscheck alerts and add support to CDB such that the hashes could be stored in the CDB database. Then, a rule could be written to check file changes against that database and optionally not alert.

aremai:
Sounds rather complicated. Is this something that could be integrated permanently into OSSEC? Surely this is something that a lot of users are facing. I don't think I'm the only one patching my machines ;)

ddpbsd:
When I find time I'm hoping to add something to analysisd to allow a user
to check a cdb before creating a syscheck alert. Creating the db will be up
to the user though.

PR #2283, opened 2026-08-05
Add per-agent FIM maintenance mode via agent_control.

atomicturtle:
Allow operators to silently accept syscheck DB updates during OS patching without stopping the manager or wiping baselines (#677, #1289, #1681).


atomicturtle:
This also closes #1793

